PR #1, opened 2021-04-26
Updated codebase to support configurable environment

farsadf:
I've replaced the hardcoded `bitclout.com` references with `environment.ts` and `environment.prod.ts`. 
By doing this, both the developers and people who want to run an instance of identity for themselves can swap in the variables and quickly build the project.

### DIFF
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -1,3 +1,8 @@
 export const environment = {
-  production: true
+  production: true,
+  hostname: 'identity.bitclout.com',
+  node_hostname: 'bitclout.com',
+  node_api_hostname: 'api.bitclout.com',
+  full_access_hostnames: ['bitclout.com', 'bitclout.green'],
+  no_access_hostnames: [''],
 };
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -1,8 +1,8 @@
 export const environment = {
   production: true,
   hostname: 'identity.bitclout.com',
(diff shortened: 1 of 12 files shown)


src/environments/environment.prod.ts
```diff
@@ -1,3 +1,8 @@
 export const environment = {
-  production: true
+  production: true,
+  hostname: 'identity.bitclout.com',
```
maebeam, line 3: I think this file is just overrides so I don't think you need to define these values twice

src/environments/environment.ts
```diff
@@ -3,7 +3,12 @@
 // The list of file replacements can be found in `angular.json`.
 
 export const environment = {
-  production: false
+  production: false,
+  hostname: 'identity.bitclout.com',
```
maebeam, line 7: perhaps name this `identityHostname`? please use camel case

maebeam (approved):

src/environments/environment.prod.ts
farsadf: I believe otherwise, the `environment.prod.ts` file will be used while building with the `--prod` flag and the other file will be used while developing locally.
For example, a developer might have configured the identity service to `identity.bitclout.me` on his local machine, while the production server uses `identity.bitclout.com`; this is a common use-case while developing locally.

src/environments/environment.ts
farsadf: Added the proper commit.

PR #3, opened 2021-04-26
More tests for IdentityService and AccountService

maebeam:


src/app/identity.service.spec.ts
```diff
@@ -1,16 +1,160 @@
...
+    //   [encryptedHex]: message,
+    // });
+  });
+
+  it('can handle info', () => {
+    // TODO
+  });
+
+  it('can handle jwt', () => {
+    // TODO
```
lazynina, line 158: We should address these todos before merging this in.

lazynina (changes requested):
Other than addressing the todos, this looks great.

Issue #4, opened 2021-04-27
Using Identity from a native mobile app


ryanjohn1:
As far as I can see, the only way to interact with the identity service such as receiving data after a successful login or to request the signing of a transaction is to post/receive a message on the window event listener. Presumably the intent is for web apps to run Identity in an iframe and communicate through that event listener. 
In a native mobile app, if we navigate the user to identity.bitclout.com in an iOS [WKWebview](https://developer.apple.com/documentation/webkit/wkwebview) or Android [CustomTab](https://developer.chrome.com/docs/android/custom-tabs/overview/) so that they have a secure environment to login then there is no way to interact with the window event listener. 
Similarly when we want to sign a transaction there is no mechanism to post an event to have identity sign the transaction hex.

Am I missing a mechanism that exists for native mobile apps to interact with Identity?

Thanks

maebeam:
You should direct the user to bitclout.com or yourbitcloutnode.com. bitclout.com embeds the identity iframe which you should be able to communicate with. If you run your own site, all it needs is an iframe with identity.

Open to ideas for making this easier if you have any

maebeam:
Actually, you might just be able to load identity.bitclout.com/embed in a WKWebView and it will support everything you need. Give that a try

ryanjohn1:
I'll check out /embed tomorrow, thanks.
But I don't think we can go with your first idea, that would be fine for
just a login flow but not for cases where we have to sign transactions e.g.
when a user tapped on "like" while scrolling their feed. Every screen would
have to be a website in order to be able to communicate with identity to
sign the transactions. It wouldn't be a native app at all.

I guess the preferred way would be an open source Identity library for each
of iOS and Android, written in Swift and Kotlin respectively. They could be
used by apps to go through the sign up/login flow using app screens instead
of web views, perform the cryptography functions and then store the data in
secure app storage rather than relying on webviews embedded within the app.

BobbyTable:
@maebeam Are there any docs on this intended usage (Iframe + message posting)? Just a sentence or three would be very helpful.

maebeam:
I think what you want is a webview of /embed and `WKWebView.evaluateJavaScript`. We want to encourage app developers to use identity.bitclout.com to log users in and sign transactions so that the user has a single place to manage all of their identities and they don't need to enter their seed phrases into random apps/websites.

maebeam:
For example: on a `WKWebView` of `/embed`, run

```
webView.evaluateJavaScript('
[...]

PR #9, opened 2021-04-29
Add support for viewing in a mobile webview

ryanjohn1:
Currently, Identity posts messages to either it's parent window or it's opener. For mobile apps that want to launch Identity in a Webview there is no way to access and respond to the messages.

To address this, I added a private function in the identity service which posts messages to the correct location. 
If an iOS or Android device registers a message handler called "appInterface" then they will be sent to that handler otherwise it will fall back to the current behaviour of using this.currentWindow which will be the opener or parent.

The checks to ensure Identity is run in either a tab or an iframe have been removed as neither is the case when a mobile webview directly accesses Identity.

Below is how it can be used, these details should be added to https://docs.bitclout.com/devs/identity-api

**Android**
An Android app can use Identity to carry out the log in flow like so:

```
webview.settings.javaScriptEnabled = true
webview.settings.domStorageEnabled = true
webview.addJavascriptInterface(MyHandler(), "bitcloutIdentityAppInterface")
webview.loadUrl("identity.bitclout.com/log-in")
```

```
class MyHandler() {
   @JavascriptInterface
   fun postMessage(data: String, transferList: String): Boolean {
      // Handle received messages
      return true
   }
}
```

Messages can be sent back to Identity with:

`webview.evaluateJavascript("window.postMessage({service: 'identity',id: 'id'})", null)`

**iOS**
An iOS app can use Identity to carry out the log in flow like so:

```
let messageHandler = MyHandler() 
let config = WKWebViewConfiguration()
config.userContentController.add(messageHandler, name: "bitcloutIdentityAppInterface")
let view = WKWebView(frame: .zero, configuration: config)
```

```
extension MyHandler: WKScriptMessageHandler {
    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
        // Handle received messages
    }
}
```

Messages can be sent back to Identity with:

`webview.evaluateJavaScript("window.postMessage({service: 'identity',id: 'id'})")`

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,12 +21,6 @@ export class AppComponent implements OnInit {
   ) { }
 
   ngOnInit(): void {
-    // We must be in an iframe OR opened with window.open
-    if (!this.globalVars.inTab && !this.globalVars.inFrame()) {
-      window.location.href = `https://${this.globalVars.environment.nodeHostname}`;
-      return;
-    }
-
     this.identityService.initialize().subscribe(res => {
       this.globalVars.hostname = res.hostname;
       if (this.globalVars.isFullAccessHostname()) {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -21,17 +21,13 @@ export class AppComponent implements OnInit {
   ) { }
 
   ngOnInit(): void {
-    this.identityService.initialize().subscribe(res => {
-      this.globalVars.hostname = res.hostname;
-      if (this.globalVars.isFullAccessHostname()) {
-        this.globalVars.accessLevelRequest = AccessLevel.Full;
-      }
-
-      this.loading = false;
-    });
-
-    // Store testnet for duration of this session
     this.activatedRoute.queryParams.subscribe(params => {
+      // We must be in an iframe OR opened with window.open OR running in a webview
+      if (!params.webview && !this.globalVars.inTab && !this.globalVars.inFrame()) {
+        window.location.href = `https://${this.globalVars.environment.nodeHostname}`;
+        return;
+      }
+      // Store testnet for duration of this session
       if (params.testnet) {
         this.globalVars.network = Network.testnet;
       }
@@ -39,6 +35,15 @@ export class AppComponent implements OnInit {
       if (params.accessLevelRequest) {
         this.globalVars.accessLevelRequest = parseInt(params.accessLevelRequest, 10);
       }
+    });
+
+    this.identityService.initialize().subscribe(res => {
+      this.globalVars.hostname = res.hostname;
+      if (this.globalVars.isFullAccessHostname()) {
+        this.globalVars.accessLevelRequest = AccessLevel.Full;
+      }
+
+      this.loading = false;
     });
   }
 }
--- a/src/app/global-vars.service.ts
+++ b/src/app/global-vars.service.ts
@@ -13,8 +13,6 @@ export class GlobalVarsService {
   hostname = '';
   accessLevelRequest = AccessLevel.ApproveAll;
 
-  inTab = !!window.opener;
-
   constructor() { }
 
   isFullAccessHostname(): boolean {
--- a/src/app/global-vars.service.ts
+++ b/src/app/global-vars.service.ts
@@ -13,6 +13,8 @@ export class GlobalVarsService {
   hostname = '';
   accessLevelRequest = AccessLevel.ApproveAll;
 
+  inTab = !!window.opener;
+
   constructor() { }
 
   isFullAccessHostname(): boolean {
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -290,12 +290,12 @@ export class IdentityService {
 
   // Post message to correct client
   private postMessage(message: any): void {
-    // iOS Webview with registered "appInterface" handler
-    if (this.currentWindow.webkit?.messageHandlers?.appInterface !== undefined) {
-      this.currentWindow.webkit.messageHandlers.appInterface.postMessage(message, '*');
-      // Android Webview with registered "appInterface" handler
-    } else if (this.currentWindow.appInterface !== undefined) {
-      this.currentWindow.appInterface.postMessage(JSON.stringify(message), '*');
+    // iOS Webview with registered "bitcloutIdentityAppInterface" handler
+    if (this.currentWindow.webkit?.messageHandlers?.bitcloutIdentityAppInterface !== undefined) {
+      this.currentWindow.webkit.messageHandlers.bitcloutIdentityAppInterface.postMessage(message, '*');
+      // Android Webview with registered "bitcloutIdentityAppInterface" handler
+    } else if (this.currentWindow.bitcloutIdentityAppInterface !== undefined) {
+      this.currentWindow.bitcloutIdentityAppInterface.postMessage(JSON.stringify(message), '*');
       // Web
     } else {
       this.currentWindow.postMessage(message, '*');
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -259,32 +259,46 @@ export class IdentityService {
     const subject = new Subject();
     this.outboundRequests[id] = subject;
 
-    this.currentWindow.postMessage({
+    this.postMessage({
       id,
       service: 'identity',
       method,
       payload,
-    }, '*');
+    });
 
     return subject;
   }
 
   // Respond to a received message
   private respond(id: string, payload: any): void {
-    this.currentWindow.postMessage({
+    this.postMessage({
       id,
       service: 'identity',
       payload
-    }, '*');
+    });
   }
 
   // Transmit a message without expecting a response
   private cast(method: string, payload?: any): void {
-    this.currentWindow.postMessage({
+    this.postMessage({
       id: null,
       service: 'identity',
       method,
       payload,
-    }, '*');
+    });
+  }
+
+  // Post message to correct client
+  private postMessage(message: any): void {
+    // iOS Webview with registered "appInterface" handler
+    if (this.currentWindow.webkit?.messageHandlers?.appInterface !== undefined) {
+      this.currentWindow.webkit.messageHandlers.appInterface.postMessage(message, '*');
+      // Android Webview with registered "appInterface" handler
+    } else if (this.currentWindow.appInterface !== undefined) {
+      this.currentWindow.appInterface.postMessage(JSON.stringify(message), '*');
+      // Web
+    } else {
+      this.currentWindow.postMessage(message, '*');
+    }
   }
 }
--- a/src/app/log-in/log-in.component.html
+++ b/src/app/log-in/log-in.component.html
@@ -1,6 +1,6 @@
 <app-banner></app-banner>
 
-<div class="container home-container" *ngIf="globalVars.inTab">
+<div class="container home-container">
   <div *ngIf="showLoadAccount" class="mb-30px">
     <div class="fs-24px font-weight-bold">
       Load your account
--- a/src/app/sign-up/sign-up.component.html
+++ b/src/app/sign-up/sign-up.component.html
@@ -1,6 +1,6 @@
 <app-banner></app-banner>
 
-<div class="container home-container" *ngIf="globalVars.inTab">
+<div class="container home-container">
   <!-- Step 1 -->
   <div *ngIf="stepNum == 1">
     <div class="fs-15px fc-muted">STEP 1 of 4</div>


andyboyd:
🎉

maebeam:
This is so cool! Looks great to me.

chrismaddern:
This looks awesome, and building support into the identity service definitely simplifies building it into an app (one less layer of message forwarding 😍 )

I think it's worth considering using a more specific namespace on the messageHandlers to prevent these messages being sent to generic bridges (particularly because of how easy it is to inject global bridge config now on iOS) e.g. `this.currentWindow.webkit?.messageHandlers?.bitCloutIdentityAppBridge...`

In the meantime, it's worth noting that this change isn't _required_ in order to use this in a WKWebView on iOS (I haven't tried on Android though).

WKWebView configuration allows for a shared process pool, and then you can implement & become the `WKUIDelegate` and handle`webView: createWebViewWithConfiguration:(WKWebViewConfiguration *)configuration forNavigationAction:....` delegate method by creating a new webview with the passed configuration and displaying it to the user.

This will maintain the pointer to the opened window and allow window <-> window communication as in a browser (this is for example, ~how Safari for iOS implements it)

e.g. 
CleanShot 2021-04-29 at 20 20 50@2x

ryanjohn1:
@maebeam Awesome, I'm glad you like it. Thanks for taking the time to look at the issue and this PR as I know you must be busy pushing out updates across what seems like a billion projects! Are there other people who need to review? 

@chrismaddern Thanks for the info, that's good to know. And I don't really see a disadvantage to making the message handler name more specific to avoid the chance of a conflict so I'll commit an update to that now

ryanjohn1:
Updated the name of the message handler and I edited my comment above to update the needed code for Android and iOS so it's clear.

chrismaddern:
Bystander L(GREAT)TM.

Hope this makes it in!

maebeam:
@ryanjohn1 I've been noodling on this and I actually want to discourage iOS apps from loading identity.bitclout.com in a `WKWebView`. Loading identity in a `WKWebView` gives the app full access to `localStorage` which is non-desirable. The iOS app should instead load identity.bitclout.com in an iframe just like the bitclout.com angular app.

Now, my iOS knowledge is not sufficient to recommend how you go about implementing this. Is it possible to create a very tiny, local WKWebView HTML and JS combo that embeds identity as an iframe and forwards messages to/from the iframe/window.open window? If not, perhaps what we actually need is an entirely new repository called mobile-identity.bitclout.com service which is a very simple, blank page angular app that forwards messages to/from WKWebView (and the android equivalent).

Maybe @chrismaddern or @andyboyd can help chime in on implementation approaches here. Looking forward to figuring this out with you all!

andyboyd:
I think one thing worth noting is that WKWebView doesn't give full access to localStorage. It allows the app to see what kind of data the browser is storing, but doesn't actually give it access to the data itself. At least as far as I can tell. 

@chrismaddern, do you know anything about WKWebView that disagrees with that?

maebeam:
Understanding the security implications of loading identity in a WKWebView would be very helpful. If it only lets you use postMessage it's great. If it lets you execute arbitrary javascript it's game over.

chrismaddern:
> Is it possible to create a very tiny, local WKWebView HTML and JS combo that embeds identity as an iframe and forwards messages to/from the iframe/window.open window

This is actually how Cloutie App works, hosted on a local server on localhost to make sure that the app never needs to load an asset on a separate server.

We could definitely separate out that code and turn it into a `BitCloutWallet` iOS dependency for anyone to use.

> @chrismaddern, do you know anything about WKWebView that disagrees with that?
Technically WKWebView allows inserting

andyboyd:
Hmm, I guess it does do that. Not being a web dev, I don't see how that's any different from what a web app can do though? Does an iframe prevent that?

I wasn't able to get it to load in an iframe inside WKWebView, so actual pointers on how to do that would be appreciated.

chrismaddern:
> Is it possible to create a very tiny, local WKWebView HTML and JS combo that embeds identity as an iframe and forwards messages to/from the iframe/window.open window

This is actually how Cloutie App works, hosted on a local server on localhost to make sure that the app never needs to load an asset on a separate server.

We could definitely separate out that code and turn it into a `BitCloutWallet` iOS dependency for anyone to use.

> @chrismaddern, do you know anything about WKWebView that disagrees with that?

WKWebView presents an interface consistent with a browser, with a couple caveats.

Technically WKWebView doesn't give a native interface to that data, but allows inserting of JS into not just the host view, but also any iframe. 

So, if we don't want to allow for `identity.bitclout.com` LocalStorage to be accessible from native bridges directly, then we need to go the _opposite_ way and make sure that there are NO injected bridges (e.g. on iOS, no `window.webkit` object... ), but then could also provide a helpful wrapper to allow for simplification of using it in an iframe in a WKWebView / Android webview.

Probably should have thought of this earlier, but for now, maintaining the need for the WKWebView → Iframe comms is a security pro as it prevents direct window access to the WKWebView you have a handle to.

andyboyd:
Frankly, I have to be honest, any approach to this that requires the use of a web app leveraging browser storage to handle security is extremely suboptimal for mobile. It would be much better to have an official library that mobile apps could use in order to sign in securely. I'd be very happy to help work on that if it was a possibility.

chrismaddern:
We could totally do that — (in fact, most of the math for it is relatively straightforward), however it comes with one downside... it means that the user has to share their raw seed phrase.

I think there's another option in the mid-term...
1. the "hard" way — implementing directly through the bridge etc...
2. an oauth approach — navigate out to a native browser for auth and support a deep link back with a bearer token

2 will require new functionality in identity.bitclout.com but would make it easier for folks to adopt if they don't want to write complex multi-layer auth schemes

andyboyd:
IMO the OAuth approach would be perfect. That's typically the way native apps interact with 3rd party services that require authentication, and is really easy to use. If we could work in that, it would be perfect.

maebeam:
> Frankly, I have to be honest, any approach to this that requires the use of a web app leveraging browser storage to handle security is extremely suboptimal for mobile. It would be much better to have an official library that mobile apps could use in order to sign in securely. I'd be very happy to help work on that if it was a possibility.

Is this true? Obviously it's not ideal but let's imagine we have a WKWebView with my proposed angular single page app at https://mobile-identity.bitclout.com that hosts an identity iframe and forwards messages to/from an iOS app. An iOS app can inject arbitrary javascript into mobile-identity.bitclout.com but can it reach inside the iFrame? If it can reach inside the iframe then there's no point in setting up https://mobile-identity.bitclout.com and we should just allow WKWebView to embed https://identity.bitclout.com.

We would absolutely love a BitClout Identity iOS app that handle key material storage/authentication/approval flows. A bounty on building this can be discussed if anyone is interested.

ryanjohn1:
If we can figure out a flow for OAuth that would be ideal. Every app I've
ever worked on has used OAuth for security whether with external services
like Google sign in or with the companies own flow. All mobile engineers
are familiar with it.
I'm struggling to visualise the flow (It's 4am for me though and I've just
woken up, so that might be it). You navigate to identity in a normal
browser, log in and it stores the private info in local storage, just like
it would on the web. You get a bearer token back along with the public
info. When I want to sign a transaction that doesn't need approval what do
I do? Normally you'd use the token to confirm to a server you are who you
are but the only place that has your info is that browser storage. You
can't access it with an https request to return a signed transaction hex.

An iOS Identity app could be created and you go through the login flow in
app. It stores the private info in its own storage which could be very
secure. But how would a client app communicate with it to get it to sign
your transaction?
On Android that would work as you can use Intents to communicate between
the two and it could do all the work and just return the public info to a
client app. But I don't know of a way of doing that on iOS?

I don't know the answer about whether you can run JavaScript against an
iframe on a page you're viewing.

wessmith:
> An iOS app can inject arbitrary javascript into mobile-identity.bitclout.com but can it reach inside the iFrame?

WKWebView _can_ inject JS into an iframe ([ref](https://developer.apple.com/documentation/webkit/wkuserscript/3656354-init)) however doing so isn't necessary. Embedding https://identity.bitclout.com in a page and handling window.open using WKWebView apis (as @chrismaddern mentioned above) is sufficient to set up the necessary communication between the iframe and the opened window.

andyboyd:
Actually, thinking about it, an official identity app may be exactly what's needed.

I'll need to have a think about the specifics of how it would work, but I'd imagine something along the lines of:

- using a URL scheme for other apps to communicate with it in order to open it up and tell it what needs done
- using the clipboard to carry the required information back to the other app

The seed phrase would need entered once per account, and then the private info for that account would be stored securely inside the identity app, where nothing else could touch it. The unencrypted seed phrase wouldn't be stored anywhere.

The big question mark for me is how would the other apps sign transactions that don't require approval? It's not a great flow if the user has to switch to another app every time they want to like a cloud. Perhaps there could also be a library the app could include that could handle that logic?

maebeam:
If we could figure out a way to make iOS cross-app communication seamless for transactions that don't require approval like likes/follows/etc... any ideas anyone?

ryanjohn1:
On iOS I think it would probably have to be a Swift library that incorporated the sign up and login screens. The private user info could be stored far more securely in the keychain than in a browser. Signing transactions etc. can be performed by the library. The library would be open source so everyone could see it's doing the same process as the original Identity library. 
The downside is that an app could access this stored info if it chose to so a potential user would still have to vet an app before deciding to use it and apps should get a security audit performed to show they're using the library appropriately. 

Users could be given a second option of how to use the library if they have an account that they mostly use to store coin with infrequent transactions. They could choose for the library not to store the private user info. Whenever they perform a transaction in the app using the library, the app would pass the transaction hex to be signed to the library. The library would show a screen for the user to enter seed phrase, it would then return the signed transaction hex to the app but not store any of the private user info. This way the app would have no way of accessing the info. 

The same thing could be created for Android too but as I said before there may be a way of creating it as an app that can sign transactions for other apps. 

@maebeam, if this is the direction you'd like to go then @andyboyd and I may be able to help. We both work for the same company, our team has been working on an Android and iOS app built on top of bitclout. We're closing in on an alpha release but a robust solution to mobile security is really important for us to get right. Some info on the company is at www.flickapp.com. Let me know what you think.

maebeam:
@ryanjohn1 can you email me so we can exchange contact info? maebeam0@gmail.com

RHatem:
@ryanjohn1  Thank you for this solution regarding mobile apps. Could you also please add support for React Native WebView.
For that, we just need to add another if condition and check if:

this.currentWindow.ReactNativeWebView !== undefined

This interface provide the postMessage() method as usual. Thank you in advance for that!

maebeam:
Did some research and I have two proposals:

1) In an ideal world we can use a `SFSafariViewController` to log the user in and sign transactions. This controller has access to cookies and local storage from safari and does not allow for arbitrary javascript injection. So, when a user clicks "Sign In with BitClout Identity", they see all their accounts and can login easily and securely. The only potential hiccup here is Apple's docs say:

> This view controller must be used to visibly present information to users; the controller may not be hidden or obscured by other views or layers. Additionally, an app may not use SFSafariViewController to track users without their knowledge and consent.

So, if we can't keep the safari view controller loaded in the background, we could also try a `ASWebAuthenticationSession`. However, this will only work if we can figure out a way to not display the modal on every transaction. If both of these options require us to interrupt the user's flow when doing simple transactions like follows, likes, etc, then we probably need Option 2.

2) We build a separate BitClout Identity app which stores all key material and runs a GCDWebServer. The web server is allowed to run in the background for up to 10 minutes. If the web server isn't running for any reason then the client calls `bitcloutidentity://sign?callback=client://`, the identity app opens, instantly redirects, and the user can continue to use the client for up to 10 minutes (or maybe more if we can enable background refresh or something).

---

Option 1 is the most ideal because it doesn't require a user to download a second app. Option 2 might make more sense because it creates a single place for users to sign in/out and revoke access to authorized apps.

Curious to hear everyones thoughts on these two ideas. We might need a quick POC / more knowledge of apple's review policies to make a decision

maebeam:
Alright, after some more discussion, sounds like Apple will reject an app that uses option 1 and we need to build option 2.

@ryanjohn1 is this something flick wants to take on? happy to discuss more over email

HPaulson:
Just a heads up to anyone trying to build a native application, this isn't *actually* necessary. Native applications (Tested on both iOS and Android) can access and receive information from Identity just fine in its current state. As @maebeam mentions [here](https://github.com/bitclout/identity/pull/9#issuecomment-830449282), giving access to locate storage is undesirable, however, this is exactly what the current method does- thus these changes should be implemented to not only make mobile access easier but also to restrict access to storage as applications are able to do now. 

Though using Identity natively on mobile is a tad bit annoying, it is totally possible- just in case anyone was waiting on this PR to implement. :tada:

maebeam:
Yes, I would love to merge some version of this PR so that mobile apps can use identity as-is while we developer an identity app

HPaulson:
> Yes, I would love to merge some version of this PR so that mobile apps can use identity as-is while we developer an identity app

To clarify my comment, Mobile apps *already can* use identity as-is- the only issue is that they have access to more data than ideal. As of right now, however, Identity does work 110% on native mobile applications, it's just a tiny bit tricky to implement (Especially for iOS).

HPaulson:
Also a quick side note, I think this PR should be marked to close #4 :)

chapel:
As far as I know, there is no way to have identity work as a client-side only browser app for signing and keep the seed away from native mobile apps. WKWebview allows full control over the browser, yes you can't read the data directly but you can inject any JS you want and even overwrite core JS libraries.

As stated by others, OAuth is usually the way this is handled, but that requires a trusted third-party server that handles whatever authentication is needed. This would not be possible for identity unless you were to actually provide an OAuth flow that was handled server-side.

With crypto projects on mobile, you have to trust the app with your seed unless you use a hardware wallet that signs transactions using BT or some other means.

I honestly feel that we should consult with a security expert on this matter as anything we come up with will likely be substandard. The balance is between ease of use and security, I don't see only one option fitting both but instead being worse for both. We should support a range of options from ease of use like many people are used to with other sites and services and more secure options that others expect.

maebeam:
@chapel please see my comment above about using GCDWebServer to sign. it solves all these problems

RHatem:
Hello together,

After doing some thinking about a mobile app solution, I found a simple one for now and would like to get your opinion on it because I am not 100% sure that it is secure.

We can store an encrypted version of the seed phrase in the memory in a private variable and try to make it as safe as possible that the application cannot inject JavaScript to access it. This means the user inputs the seed phrase once at starting the app and has to input it again every time if the app is closed completely (also from the background). We will encrypt the seed phrase with a random generated key that is also stored somewhere in the memory. This add another security layer.
Let me know if this solution is secure for now.

If you think that is the case, I will be happy to work on one.

chapel:
> @chapel please see my comment above about using GCDWebServer to sign. it solves all these problems

Fair enough, I think I misread a bit of what you said. I am not so sure it is a slam dunk solution, as you brought up iOS killing backgrounded apps commonly, well at my company it is a serious issue that we have to deal with all the time. Also what about Android?

I guess I just feel like there is no perfect solution and this is not solved for any crypto projects that I know of. Security and ease of use are opposing forces and very hard to come together and make both sides happy.

ryanjohn1:
I've chatted to the iOS engineers here and they also think that a GCDWebServer solution probably won't work very well, as @chapel says iOS aggressively kills backgrounded apps processes and is quite unpredictable. There are also cases like low power mode where background processes are disabled and bitclout apps wouldn't work.

My view is that we can't have the assumption that a user can put their seed phrase into an app and not trust the app. The only way to do that would be to hold the seeds on a central server which signed transactions. But this would make it a huge hacking target and would be putting everyones eggs in one basket. 

As @HPaulson edited post above shows as well as other info in this thread, storing private user info in local browser storage isn't secure. People have found ways around it for building useful things, it won't work in the long term to prevent bad actors from stealing seeds.

With apps, seeds can be stored very securely but trust will have to be placed in the app by the user. You need to know that they're not a scam before entering your seed phrase. Security audits can help prove that an app is legitimate as well as a proven public record as a company. 

Separately it might be good to come up with an easy way for people to prove to a website/app that they are the owner of an account which can be used for proving coin ownership or similar cases without entering a seed phrase.

My view on next steps:
1. I think this PR is fine to merge as it doesn't change anything except make it easier for mobile developers to integrate with Identity
2. Create native iOS and Android libraries to store seed phrases securely within the app and avoid the browser all together. Users could then trust a client app they trust which uses the library.
3. Today I'm going to investigate whether or not it's possible to create an Android Identity app which would allow users to just trust that one app, which would handle log-in views as well as seamless transaction signing. If that does work then it could be the most secure solution for Android rather than a library.

HPaulson:
@ryanjohn1 I'm unsure if this is the solution, however I don't have a better viable option at this point unfortunately. 

Being forced to trust an application with your seed, in my mind, defeats the purpose of having identity in the first place. In this case, isn't it the same auth flow as we have prior to the release? 

Unfortunately, however, I don't have a more viable solution at this point, and I don't want to disagree with a solution without providing a replacement. I do think though that we should consider the security aspects here, as the solution above only handles the usability side of the scale imho.

ryanjohn1:
I can confirm that on Android we could create an Identity app which the user could download from the store. They could sign up/log in the app and their seed info would be stored securely in the app. Then any bitclout client apps they download could request to sign a transaction etc. through a message to a service of the Identity app. It could then calculate the transaction hex using the stored private info and then return the new transaction hex through the same message system. It would basically run like Identity on the web but with more secure storage and you'd download the app first from the store. Apps could link to the store page for Identity for users who come to their app first without setting up Identity.
I just put together a quick POC which just tested out the messaging part.
It's using the Messenger concepts found here:
https://developer.android.com/guide/components/bound-services#Messenger
But expanded to handle two way communication (responding back to the client)

maebeam:
I think we don't have any option but to use GCDWebServer. We cannot expect users to know which apps have been audited and we do not want to train users to put their seed phrases anywhere but Identity. 

The app will still work in low power mode, it will fall back to using URI redirects. There can be an in-app banner that says low power mode doesn't work well with bitclout. 

Before we dismiss GCDWebServer I'd prefer to build a very cheap POC and see how well it works as the security advantages are incredibly compelling.

maebeam:
The only other option is to ask Apple for an exception to keep a safari view controller alive in the background after logging in, but it seems incredibly unlikely they'd make an exception for us.

chapel:
@maebeam I fully agree we should test our assumptions, I just worry it won't solve all the issues like we wish it would. If someone can get it working then fantastic!

I'm not well versed in writing blockchains, so this idea might not be possible whatsoever, but what about adding the concept of sub-accounts, which a user can create giving a subset of permissions (like follow/like/post) to act as the primary account. This could enable a safer way for users to log into other apps without risking their coins and only use apps they trust for permission to deal with the monetary side. Could also enable a way to revoke those sub-accounts, which would mean users could stop access if the seed got exposed.

Additionally, this would allow sharing of accounts by teams without exposing the money on the account.

andyboyd:
Thinking about it, we just need to use URL schemes to communicate with a native Identity app. That app would do exactly what the identity web service does. It could use a WKWebView to just wrap the identity web service, or it could implement the same logic in native code, whichever is easier.

The app would have a known URL scheme, which bitclout client appscould use to send it requests. Each request would need to be given a redirect URL. The client is responsible for having its own custom url scheme, so that the app can send it back a response, as well as generating its own unique encryption seed for each request.

The Identity app would then receive those requests, do what it needs to, and send the response (which would be either the login public info, or the signed transaction hex) back to the original app via the redirect URL scheme. It could check that the scheme was not `http`, `https` or similar, to prevent those redirect URLs actually going to a server somewhere.

The original app would then pick that up, and would be responsible for taking it from there to confirm the transaction.

How does that sound? No need for background web servers, or secret invisible browser windows.

chrismaddern:
Quick drive-by comment... (work has left less time for BitClout fun this past week :( )

I'm think both of these solutions may run into Apple approval challenges (as it happens, we've explored both before at Button, where app → apps comms is a component of our Brand tech stack.) 

## Required App → App switch for auth
Apple explicitly forbids an app from _requiring_ another app in order to be able to function in the [App Store Guidelines](https://developer.apple.com/app-store/review/guidelines/).
Screenshot 2021-05-03 at 13 25 01@2x

Of the two, this seems like the most likely to be able to get an exception on, but importantly, we wouldn't need _one_ exception, you'd need one for each app that _used_ the identity wallet.

FWIW, also strongly recommend using Universal Links (https://) vs. scheme based links as there is no scheme registry and any app could take over either side of that two way comms simply by registering the scheme. The operating system enforces the domain<>app package trust relationship for Universal Links / App Links on Android.

## Offscreen Safari VC
There are both technical & approval challenges to this — the component is explicitly designed to ensure that this doesn't happen so even if you got the exception (again, one for every app, not just Identity) I would expect it to be tough to implement. They tend to become unresponsive / release offscreen.

## General thoughts on GCDWebServer
(This will also run into the "requiring another app to function" review challenge. )

I am however, personally really interested to know whether it would work now (last time we tried, the ports were inaccessible quite quickly when the app backgrounded, but the backgrounding model has changed significantly since then and allows much more background activity, including explicitly several activities that may require port exposure)

I don't think there's going to be a perfect solution here on iOS — the entire iOS sandboxing system is designed to stop XPC outside of very limited use-cases, and the guidelines prevent requiring another app (which it would be very hard to communicate with due to #1 🤣 ) 

## Longer term — Secure Enclave?
It's not a today solution, but longer term, the device Secure Enclave presents a perfect solution to on-device signing use-cases — it exists to perform cryptographic operations on input, sharing only the output and is completely opaque to apps (infact, all _software_)

Today, there are two main challenges:
* No support for secp256k1
* You can't add an existing private key, only generate a new one, and then can't get that pKey out

I think that in a few years, the importance of crypto wallets will demand a device / OS level solution. I think it's worth writing this up as a RADAR and submitting it to Apple to get it on their radar (oh, I finally understand the name!) and could maybe burn a support ticket to see if they have any near term ideas.

## Android
As someone indicated above, you can totally do this with XPC Messages on Android where application sandboxing is much more permissive, or run a web server.



I'll keep thinking about alternatives, but in the meantime, it sounds like we can use the existing solution for iOS apps in the interim.

andyboyd:
That'll depend on whether the identity app can send back enough information on the original login request to be able to sign transactions that don't require approval.

If the primary concern is the private info, and it being stored somewhere that a 3rd party app could potentially access, we could also use a client library that apps would need to use in order to communicate with the identity app. This library could add an additional layer of encryption on top of whatever gets stored inside the app sandbox, and it could be distributed in binary form so it could keep whatever encryption seeds it uses secret.

It would basically be like two way SSL pinning, except instead of encrypting network traffic, it encrypts traffic between the identity app and the client library

maebeam:
@chrismaddern thank you for your insight and for pointing out rule 4.2.3. This is a huge blow to the "require an identity app" path.

I think @chapel might be on to something about deriving keys that are allowed to only sign likes, follows, etc. This would require a consensus-level change but it might be worthwhile given the roadblocks we're encountering on iOS.

maebeam:
Following up on 4.2.3, this post from two years ago says only part of the app needs to function without the identity app. We could allow users to browse logged out but signing in/up could require the identity app https://developer.apple.com/forums/thread/114795

If this is true, then I am still most in favor of attempting to use GCDWebServer

maebeam:
> That'll depend on whether the identity app can send back enough information on the original login request to be able to sign transactions that don't require approval.

For this to be the case we would need a derived key and consensus level changes.

> If the primary concern is the private info, and it being stored somewhere that a 3rd party app could potentially access, we could also use a client library that apps would need to use in order to communicate with the identity app. This library could add an additional layer of encryption on top of whatever gets stored inside the app sandbox, and it could be distributed in binary form so it could keep whatever encryption seeds it uses secret.

Security through obscurity is not sufficient here. We need open source libraries that everyone agrees would never allow a third party app to bypass the approval flow when it's required.

ryanjohn1:
I'll admit I don't understand a huge amount about the blockchain data
structure yet but I do think something that makes a clear distinction
between monetary based transactions and purely social interactions would be
hugely beneficial. Beyond the access level in Identity, something that is
baked deeper in

chrismaddern:
I _love_ the idea of derived keys with authority only for signing certain types of transactions 😍 , however am also very aware that it's not me who has to implement it!

There's a tension at the very core of social-meets-financial (we felt it acutely in building Venmo), and when you decentralize that, you can't easily rely on the centralized authority to make new policy. Building early support into the protocol for permissions sounds smart, beyond just the iOS login problem.

e.g. What happens when BitClout is a primary social media platform, and companies want their social media teams to be out there managing the account? Now we need to protect not just against a potential developer, but also an agent acting on your behalf who we probably don't want to give either unfettered signing ability OR access to the seed phrase.

chapel:
Some great insights by all here.

In addition to the derived keys, we will definitely need multisig support which could be another way of protecting accounts. Having financial transactions require multisig but social not.

Maybe it is worth creating a bare-bones identity app to see if we can get it past Apple's review, if they allow it and the UX is good enough then it can be invested in further to improve the mobile experience, thinking about notifications or popups to approve/deny certain transactions. Bare-bones as others have said, would just be wrapping identity.bitclout.com and just enabling the cross-app communication.

src/app/app.component.ts
```diff
@@ -21,12 +21,6 @@ export class AppComponent implements OnInit {
   ) { }
 
   ngOnInit(): void {
-    // We must be in an iframe OR opened with window.open
```
maebeam, line 24: So, with this change, a user can visit identity.bitclout.com but when they try to sign up or log in their browser nothing will happen.

Could I request that you either add support for calling `window.open(bitclout.com)` when the user completes a log in or sign up flow OR disable the redirect if some param like `?webview=true` is present?

maebeam (changes requested):

src/app/app.component.ts
maebeam: Once this is resolved I am happy to merge away

HPaulson (approved):
This LGTM now... while it's not a perfect solution, we're definitely moving in the right direction. Don't forget to add "close #4" to the PR description :)

ryanjohn1:
Rebased to fix the conflicts. 
I can't manually link the issue as I don't have write access, I'm not seeing a way of doing it without committing something else so whoever merges it can hopefully link the issue quickly using the UI top right.

@RHatem Sorry, I had a look at adding support for React Native webviews but without a handy project to test it on I didn't want to put some speculative code in for it. It might be best if you raise a PR with that and I imagine that won't be very controversial

src/app/app.component.ts
```diff
@@ -21,23 +21,13 @@ export class AppComponent implements OnInit {
...
-        this.globalVars.accessLevelRequest = AccessLevel.Full;
-      }
-
-      this.loading = false;
-    });
-
-    // Store testnet for duration of this session
     this.activatedRoute.queryParams.subscribe(params => {
+      // We must be in an iframe OR opened with window.open OR running in a webview
+      if (!params.webview && !this.globalVars.inTab && !this.globalVars.inFrame()) {
```
maebeam, line 26: Did you test this locally? Angular is very annoying and fires queryParams twice and the first event is always empty. This might redirect regardless of the webview param?

ryanjohn1: Yeah, I tested it and it worked fine. If I pass webview=true it didn't redirect. If I left it off it redirected, so I think it's fine

addanus:
Sorry. Bit L8 to the party, but had to pondr...

* for the mobile login (and web)
    A. the user launches a client app and needs to login/make a transaction
    B. client app sets up a login/transaction request, signs it using prvkey and drops it on the blockchain
    ... for logins, a token timeout is requested (and the processor can override/reject it if malformed/too long)
    C. the client app then launchs a token processor (app) via deep link - giving transaction id/etc
    D. the processor pulls the transaction from the blockchain and confirms/etc, then pulls details by using the pubkey to decrypt
    E. the processor asks the user to sign the request - which they can do using their secret/etc
    ... it can show details like what ops can be done and for $ transactions, how much can be transfered theoretically
    ... in fact, $ transactions should require a 1 time approval - and so the client can simply ask the user to buy their coin
    F. when all good, the processor pushes an approved auth/payment transaction to the blockchain and sends the user back to the client
    G. the client uses the auth token (if this was a login) to make future requests until the token expires
    H. the (nodes) validate that the token is valid by decrypting some details using the client's pubkey and the orig node's pubkey


* the result is it doesn't matter that the client/processor goes to sleep while the other's awake
* all this could be done using a smart contract of course - and every actor (except the user?) can be open sourced
* the processor/other app can monitor the blockchain on the user's behalf and send notifications if smt out of the ordinary happens
* other processors can come along and the user can choose a prefered one to handle bitclout auth/transaction requests

Thoughts?

maebeam:
I fixed some things and merged #16

HPaulson:
@maebeam Could you please enable GitHub Discussions? I think it'd be an appropriate place to continue the conversation regarding Identity on Native devices

PR #17, opened 2021-05-06
Fix mobile webviews as query params weren't being applied 

### LINKED ISSUE #4: Using Identity from a native mobile app

ryanjohn1:
After the previous PR some changes were made to make more use of the webview query param to disable the banner and add an extra check to the message posting. Unfortunately this broke things as the initialize event was sent out immediately before the query params have come through, so the identity service thinks its not running in a webview and posts the message to the standard window. I've removed that check as the presence of a "bitcloutIdentityAppInterface" is enough to be sure that the loader of this page has opted in to having the message sent outside of the usual window.

The other issue was due to the ordering of the initial calls in app.component so I've rearranged them to set the globalVars first before checking the redirect as this was checking for webview in globalVars before it was set

Closes #4

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -3,6 +3,7 @@ import {ActivatedRoute} from '@angular/router';
 import {GlobalVarsService} from './global-vars.service';
 import {IdentityService} from './identity.service';
 import {AccessLevel, Network} from '../types/identity';
+import {skip} from 'rxjs/operators';
 
 @Component({
   selector: 'app-root',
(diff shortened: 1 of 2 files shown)


maebeam:
Fixed in #18

Issue #24, opened 2021-05-25
Identity sometimes asks for approval when decrypting even with access level 4


andyboyd:
Not sure under what circumstances this happens, but it seems to be related to how long it's been since access was originally granted. But sometimes when sending a `decrypt` message to identity, it's returning `approvalRequired: true`, even though `AccessLevel: Full` was requested and granted. This shouldn't be happening, full access should mean full access, and approval should not be required for anything. 

I'm using this in a native mobile app with URL: `"https://identity.bitclout.com/embed?accessLevelRequest=4&webview=true"` and the decrypt message being sent is:

```
{
  id: <some uuid>,
  service: 'identity',
  method: 'decrypt',
  payload: {
    accessLevel: 4,
    accessLevelHmac: <stored hmac from login>,
    encryptedSeedHex: <stored encrypted hex from login>,
    encryptedHexes: [<array of encrypted hexes>],
  },
}
```

andyboyd:
@maebeam , any input on this? Would be good to know if this behaviour is expected, because it's pretty breaking right now for us.

maebeam:
Sorry I missed this, for some reason my notifications were misconfigured.

Safari expires third party cookies after one week. If the cookie is expired we need the user to visit identity and click a button so we can store the cookie for another week. It sounds like we may need a special case where we use localStorage for Safari first party webviews embedded in apps

maebeam:
Should be as simple as changing this function:

```
  mustUseStorageAccess(): boolean {
    return typeof document.hasStorageAccess === 'function' && !this.globalVars.webview;
  }
```

maebeam:
Give that a try locally and either submit a PR or I can make the change for you

andyboyd:
I can give it a try locally, but if I understand correctly, I'll not be able to know if it's worked for a week? I think I'd prefer if you just made the change if that's ok, we've got actual real life users on our app now, so would be best if the fix got in as quickly as possible, and that'll probably happen faster with you making the changes. 

I imagine after this fix is applied, user's will need to log out and back in to make sure identity stores things in the right place? If so, I'll need to put something into the app to make that happen when it needs to.

andyboyd:
Actually, I was able to verify locally that it works by changing the system time. I'll raise a PR now. Unfortunately it does seem that users will have to log out and back in again to avoid getting stuck, but I don't think that can be helped.

andyboyd:
[...]

PR #27, opened 2021-05-28
Making native clients use local storage instead of cookies for storage

andyboyd:
Fixes Issue #24, where identity in a native app was requiring approval after 7 days, even when Full Access Level had been granted.


andyboyd:
@maebeam This is the PR you asked me to raise

maebeam:
Merged, thanks!

PR #42, opened 2021-06-20
Display a message when sending to yourself

trevormil:
Previously displayed a blank message for a basic transfer txn (i.e. whole message was just "localhost wants to send") when sender and recipient were the same.

Now displays "localhost wants to send $CLOUT to public key".


maebeam (approved):
Looks great! Thanks for this change. Will get it merged soon.

PR #45, opened 2021-06-28
Support Sign-In w/ Google Webview

thehappypenguin:
Today, Bitclout Identity doesn't support sign-in with Google from a WebView. This is because the OAuth callback from Google doesn't carry-over the `webview=true` parameter from the original Identity context.

This PR uses Google's "state" query parameter to properly send the context back to Bitclout Identity and inform it that the login attempt came from a WebView so it can initialize correctly.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -44,7 +44,7 @@ export class AppComponent implements OnInit {
       this.globalVars.webview = true;
     }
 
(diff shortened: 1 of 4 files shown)


maebeam (approved):
Looks good to me. If you're willing, it would be great to do the same thing for the `testnet` param so aren't appending it to the URL on line 63 of `log-in.component.ts`

PR #48, opened 2021-07-04
Add shared secret encryption/decryption

AeonSw4n:
Moved message encryption to identity.
`crypto.service.ts` - decode public key to EC point buffer
`identity.service.ts` - added 'encrypt' method
`signing.service.ts` - handle encryption; modified decryption to support both legacy and shared secret encryption.
`index.js` - had to add second chunk to AES, otherwise, decryption wasn't working. Legacy encryption/decryption and AES added to support old messages. Added encryptShared and decryptShared which use ECDH based KDF.

### DIFF
--- a/src/app/crypto.service.ts
+++ b/src/app/crypto.service.ts
@@ -136,6 +136,21 @@ export class CryptoService {
     return bs58check.encode(prefixAndKey);
   }
 
+  // Decode public key base58check to Buffer of secp256k1 public key
+  publicKeyBase58CheckToECBuffer(publicKeyBase58Check: string): Buffer {
+    // Sanity check similar to Base58CheckDecodePrefix from core/lib/base58.go
+    if (publicKeyBase58Check.length < 5){
+      throw new Error('Failed to decode public key');
+    }
+    const decoded = bs58check.decode(publicKeyBase58Check);
+    const payload = Uint8Array.from(decoded).slice(3);
+
+    const ec = new EC('secp256k1');
+    const publicKey = ec.keyFromPublic(payload, 'array');
+
+    return new Buffer(publicKey.getPublic('array'));
+  }
+
   keychainToBtcAddress(keychain: HDNode, network: Network): string {
     const prefix = CryptoService.PUBLIC_KEY_PREFIXES[network].bitcoin;
     // @ts-ignore TODO: add "identifier" to type definition
--- a/src/app/crypto.service.ts
+++ b/src/app/crypto.service.ts
@@ -137,18 +137,18 @@ export class CryptoService {
   }
 
   // Decode public key base58check to Buffer of secp256k1 public key
-  publicKeyBase58CheckToECBuffer(publicKeyBase58Check: string): Buffer {
+  publicKeyToECBuffer(publicKey: string): Buffer {
     // Sanity check similar to Base58CheckDecodePrefix from core/lib/base58.go
-    if (publicKeyBase58Check.length < 5){
+    if (publicKey.length < 5){
       throw new Error('Failed to decode public key');
     }
-    const decoded = bs58check.decode(publicKeyBase58Check);
+    const decoded = bs58check.decode(publicKey);
     const payload = Uint8Array.from(decoded).slice(3);
 
     const ec = new EC('secp256k1');
-    const publicKey = ec.keyFromPublic(payload, 'array');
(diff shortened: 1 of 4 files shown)


src/app/identity.service.ts
```diff
@@ -97,14 +98,28 @@ export class IdentityService {
...
+    });
+  }
+
   private handleDecrypt(data: any): void {
     if (!this.approve(data, AccessLevel.ApproveAll)) {
       return;
     }
 
-    const { id, payload: { encryptedSeedHex, encryptedHexes } } = data;
+    const { id, payload: { encryptedSeedHex, encryptedHexesAndPublicKeys } } = data;
```
maebeam, line 120: Can we maintain backwards compatibility here and support the old param?

src/app/identity.service.ts
```diff
@@ -97,14 +98,36 @@ export class IdentityService {
     });
   }
 
+  // Encrypt with shared secret
+  private handleEncrypt(data: any): void{
+    if (!this.approve(data, AccessLevel.ApproveAll)){
+      return;
+    }
+
+    const { id, payload: { encryptedSeedHex, recipientPubkey, message} } = data;
```
maebeam, line 107: nit: Can we name this param `recipientPublicKey`

src/app/signing.service.ts
```diff
@@ -21,20 +22,72 @@ export class SigningService {
     return jsonwebtoken.sign({ }, encodedPrivateKey, { algorithm: 'ES256', expiresIn: 60 * 10 });
   }
 
-  decryptMessages(seedHex: string, encryptedHexes: string[]): { [key: string]: any } {
+  encryptMessage(seedHex: string, recipientPubkeyBase58Check: string, message: string): string {
```
maebeam, line 25: I think it's fine to just call this parameter `recipientPublicKey`. Any string representation of a public key will necessarily be base58

src/app/crypto.service.ts
```diff
@@ -136,6 +136,21 @@ export class CryptoService {
     return bs58check.encode(prefixAndKey);
   }
 
+  // Decode public key base58check to Buffer of secp256k1 public key
+  publicKeyBase58CheckToECBuffer(publicKeyBase58Check: string): Buffer {
```
maebeam, line 140: All string representations are base58. So how about 
```
  publicKeyToECBuffer(publicKey: string): Buffer {
```

src/app/identity.service.ts
```diff
@@ -97,14 +98,36 @@ export class IdentityService {
...
     }
 
-    const { id, payload: { encryptedSeedHex, encryptedHexes } } = data;
-    const seedHex = this.cryptoService.decryptSeedHex(encryptedSeedHex, this.globalVars.hostname);
-    const decryptedHexes = this.signingService.decryptMessages(seedHex, encryptedHexes);
+    const seedHex = this.cryptoService.decryptSeedHex(data.payload.encryptedSeedHex, this.globalVars.hostname);
+    const id = data.id;
+
+    let decryptedHexes;
+    if (data.payload.encryptedHexes){
```
maebeam, line 124: Can we have two separate methods on signing service and call them here? E.g. `decryptMessages` and `decryptMessagesLegacy`. Will prevent us from needing `string` vs `object` checks in `decryptMessages`. Code duplication is fine because `decryptMessagesLegacy` will be deleted in a few weeks

src/lib/ecies/index.js
```diff
@@ -141,5 +162,35 @@ export const decrypt = function(privateKey, encrypted) {
...
+export const decryptShared = function(privateKeyRecipient, publicKeySender, encrypted, opts) {
+  opts = opts || {};
+  const sharedPx = derive(privateKeyRecipient, publicKeySender);
+  const sharedPrivateKey = kdf(sharedPx, 32);
+
+  opts.legacy = false;
+  return decrypt(sharedPrivateKey, encrypted, opts);
+}
+
+// @petern was here
```
maebeam, line 196: please nix this ;)

maebeam (commented):
looking very good! just a few more small changes por favor 🙏🏼

src/lib/ecies/index.js
AeonSw4n: Sure haha.

maebeam (approved):

PR #52, opened 2021-07-11
Fix message display errors by predefining private key buffer length

### LINKED ISSUE #44: Can't decrypt messages: Bad private key
Any idea how to help this user?

debug2

Onurklngc:
Fixes #44. For some accounts, the private key buffer array is created with a length of 31 instead of 32 based on its element values. I was able to view the received messages after the change.


maebeam (approved):
Phenomenal work!

AeonSw4n:
@onurklngc Thank you so much for this PR!

maebeam:
I will get this deployed soon

PR #53, opened 2021-07-15
Pn/derived keys

AeonSw4n:


BogdanDidenko:
Interesting direction
But it is not completely clear to me from the code how it should work.

Could you describe your concept in a few words?

andyboyd:
@BogdanDidenko , I suspect it's a counterpart to this: https://github.com/bitclout/core/pull/69

src/app/derive/derive.component.ts
```diff
@@ -0,0 +1,83 @@
...
+          this.allUsers[user.PublicKeyBase58Check] = {
+            username: user.ProfileEntryResponse?.Username,
+            profilePic: user.ProfileEntryResponse?.ProfilePic,
+          };
+        }
+      });
+    }
+  }
+
+  launchGoogle(): void {
```
maebeam, line 52: Can we de-dupe this code and move it into a helper somewhere?

src/app/derive/derive.component.ts
```diff
@@ -0,0 +1,83 @@
...
+    public globalVars: GlobalVarsService,
+    private backendApi: BackendAPIService,
+  ) { }
+
+  ngOnInit(): void {
+    // Load profile pictures and usernames
+    this.loadUsers();
+  }
+
+  loadUsers(): void {
```
maebeam, line 33: Let's de-dupe this code too

maebeam (commented):
looks great, let's just de-dupe

src/app/account.service.ts
```diff
@@ -71,6 +78,37 @@ export class AccountService {
     }
   }
 
+  getDerivedUser(publicKey: string, blockHeight: number): DerivedUserInfo{
+    const privateUser = this.getPrivateUsers()[publicKey];
+
+    if (!this.entropyService.temporaryEntropy.mnemonic) {
```
maebeam, line 84: If for some reason this method was called multiple times in a row (usually it shouldn't be) wouldn't the same entropy be reused? Might be a good idea to always call `setNewTemporaryEntropy`?

src/app/account.service.ts
```diff
@@ -71,6 +78,37 @@ export class AccountService {
...
+      this.entropyService.setNewTemporaryEntropy();
+    }
+    const derivedMnemonic = this.entropyService.temporaryEntropy?.mnemonic;
+    const derivedKeychain = this.cryptoService.mnemonicToKeychain(derivedMnemonic);
+    const derivedSeedHex = this.cryptoService.keychainToSeedHex(derivedKeychain);
+    const derivedPrivateKey = this.cryptoService.seedHexToPrivateKey(derivedSeedHex);
+    const derivedPublicKey = this.cryptoService.privateKeyToBitcloutPublicKey(derivedPrivateKey, privateUser.network);
+
+    // By default we authorize this derived key for 2500 blocks.
+    const expirationBlock = blockHeight + 2500;
```
maebeam, line 94: One week might be too short. I think this should probably be 2-4 weeks

src/app/backend-api.service.ts
```diff
@@ -23,7 +30,37 @@ export class BackendAPIService {
       },
     );
   }
+
+  GetUserProfiles(
```
maebeam, line 34: Great work de-duping this

src/app/backend-api.service.ts
```diff
@@ -23,7 +30,37 @@ export class BackendAPIService {
...
+    publicKeys: any[]
+  ): Observable<{[key: string]: UserProfile}> {
+    return new Observable<{[key: string]: UserProfile}>(subscriber => {
+      const userProfiles: {[key: string]: any} = {};
+      for (const publicKey of publicKeys) {
+        userProfiles[publicKey] = {};
+      }
+
+      if (publicKeys.length > 0) {
```
maebeam, line 43: else compete the observable immediately?

src/app/backend-api.service.ts
```diff
@@ -3,11 +3,18 @@ import {HttpClient} from '@angular/common/http';
...
+  profilePic: any;
+}
+
 @Injectable({
   providedIn: 'root'
 })
+
 export class BackendAPIService {
   endpoint = `https://${environment.nodeHostname}/api/v0`;
+  exchange = `https://${environment.nodeHostname}/api/v1`;
```
maebeam, line 17: This endpoint is way too slow for just fetching the tip height. We probably need a new endpoint. Or, alternatively, Lazy Nina just merged a PR that adds GetAppState to this file. Let's include the tip height in the app state.

src/app/derive/derive.component.html
```diff
@@ -0,0 +1,36 @@
+<app-banner></app-banner>
+
+<div class="container home-container" *ngIf="globalVars.inTab || globalVars.webview">
```
maebeam, line 3: I don't love that this code is duplicated. Can we just reuse the log-in component and add a `globalVars.derive` query param?

src/app/account.service.ts
AeonSw4n: Makes sense to me.

maebeam (approved):

maebeam:
Closed in favor of #71

src/app/account.service.ts
```diff
@@ -74,6 +82,42 @@ export class AccountService {
     }
   }
 
+  getDerivedUser(publicKey: string, blockHeight: number): DerivedUserInfo{
```
diamondhands0, line 85: nit: Please rename to getDerivedPrivateUser, since this returns the seed hex for the derived key.

src/app/app.component.ts
```diff
@@ -51,11 +52,21 @@ export class AppComponent implements OnInit {
     if (params.get('hideGoogle')) {
       this.globalVars.hideGoogle = true;
     }
-    
+
+    // Callback should only be used in mobile applications, where payload is passed through URL parameters.
+    if (params.get('callback')) {
+      try {
+        this.globalVars.callback = new URL(params.get('callback') as string);
```
diamondhands0, line 59: How is callback related to hostname? Do we want to enforce that the callback has the same hostname as the thing that called this? I don't think it's necessary but put some juices on it and lmk.

src/app/google-drive.service.ts
```diff
@@ -54,4 +60,27 @@ export class GoogleDriveService {
       `https://www.googleapis.com/drive/v3/files/${fileId}?${httpParams.toString()}`,
       { headers: { Authorization: `Bearer ${this.accessToken}` }});
   }
+
+  public launchGoogle(): void {
```
diamondhands0, line 64: How does this google stuff fit into everything?

src/types/identity.ts
```diff
@@ -28,6 +28,17 @@ export interface PublicUserInfo {
   version: PrivateUserVersion;
 }
 
+export interface DerivedUserInfo {
```
diamondhands0, line 31: nit: Change this to DerivedPrivateUser since this contains key material.

src/app/identity.service.ts
```diff
@@ -67,6 +72,24 @@ export class IdentityService {
...
+  }): void {
+    if (this.globalVars.callback) {
+      // If callback is passed, we redirect to it with payload as URL parameters.
+      let paramString = '?';
+      for (const key in payload.derivedUserInfo) {
+        if (payload.derivedUserInfo.hasOwnProperty(key)) {
+          paramString += key + '=' + (payload.derivedUserInfo as any)[key].toString() + '&';
+        }
+      }
+      paramString = paramString.slice(0, -1);
```
diamondhands0, line 86: nit: This is a bit clever, which is generally bad but acceptable in this case... It just so happens that there will either be a ? or an & at the end. I would add a comment explaining tihs cleverness. It takes some brianpower to figure out why this won't break in certain caises. When that's the case, always add a comment to explain the situation. This helps us avoid breaking it in the future, or at least helps us understand why it broke if we're coming back to it.

src/app/google-drive.service.ts
diamondhands0: I see.. it was just deduping. Carry on...

src/lib/bindata/util.ts
```diff
@@ -28,3 +28,20 @@ export const uvarint64ToBuf = (uint: number): Buffer => {
 
   return new Buffer(result);
 };
+
+export const uint64ToBufBigEndian = (uint: number): Buffer => {
```
diamondhands0, line 32: Where'd you get this from. Are you 100% sure it works?

src/app/derive/derive.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // Set withCallback
+    this.withCallback = this.globalVars.callback !== null;
+  }
+
+  launchGoogle(): void {
+    this.googleDrive.launchGoogle();
+  }
+
+  selectAccount(publicKey: string): void {
```
diamondhands0, line 48: nit: I don't like that this has the same name as the log-in component. Can we rename this to selectAccountAndDeriveKey?

PR #59, opened 2021-08-10
Fix access level being set incorrectly in seed flow

diamondhands0:
The access level is not being set correctly after an account is added
via the seed flow. This is causing third-party nodes to error when
opening up Jumio after the seed flow.

The error happens because we get a null user returned here:
https://github.com/bitclout/identity/blob/01f13fa46198d6273483f9e8818b63928d911aea/src/app/backend-api.service.ts#L39

The error is caused by the access level check here:
https://github.com/bitclout/identity/blob/b9c367583b1630feb3c70df27fda50600adeeb77/src/app/account.service.ts#L37

### DIFF
--- a/src/app/sign-up/sign-up.component.ts
+++ b/src/app/sign-up/sign-up.component.ts
@@ -88,6 +88,8 @@ export class SignUpComponent implements OnInit, OnDestroy {
       btcDepositAddress,
       network,
     });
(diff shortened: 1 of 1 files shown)


lazynina (approved):
:+1:

maebeam (approved):

PR #63, opened 2021-08-31
pass referral code to identity and then to the backend when beginning jumio

lazynina:


src/environments/environment.prod.ts
```diff
@@ -1,7 +1,7 @@
 export const environment = {
   production: true,
   hostname: 'identity.bitclout.com',
-  nodeHostname: 'bitclout.com',
+  nodeHostname: 'localhost:18001',
```
redpartyhat, line 4: Same comment as on the frontend PR.  Is this correct?  Seems weird to have localhost as an endpoint in the prod.ts file

src/app/backend-api.service.ts
```diff
@@ -11,7 +11,7 @@ import {GlobalVarsService} from './global-vars.service';
   providedIn: 'root'
 })
 export class BackendAPIService {
-  endpoint = `https://${environment.nodeHostname}/api/v0`;
+  endpoint = `http://${environment.nodeHostname}/api/v0`;
```
redpartyhat, line 14: I looked at the backend API endpoints and this seems fine but given that identity is very security sensitive, it seems undesirable to downgrade this to http.  Is this because localhost and https don't play nicely together?  If so, maybe we restrict this to only be http on localhost?

src/environments/environment.prod.ts
lazynina: just for testing purposes - reverted.

maebeam (approved):

src/app/jumio/jumio.component.ts
```diff
@@ -47,7 +47,10 @@ export class JumioComponent implements OnInit, OnDestroy {
   }
 
   getJumioURL(routeSuffix: string): string {
-    const url = new URL(`${window.location.origin}/${routeSuffix}`)
+    // Jumio URLs cannot be localhost, so default to bitclout.com if localhost.
+    const origin = window.location.origin.startsWith("http://localhost") ? "https://bitclout.com" : window.location.origin;
```
diamondhands0, line 51: Idk what the ramifications are of this, but I could trick this code into setting bitclout.com as the origin by registering a domain like localhostfun.com. Is there a way to make this more robust so that this isn't possible, just to eliminate any edge-cases that could cause?

lazynina: Sure we can make it more robust but all that happens if you trick this function is that you get redirected to bitclout.com instead of the local host url so I don't think it's a huge deal. This case is only on there to support running a local identity

PR #64, opened 2021-09-06
Changes "update profile" transaction to require Access Level 4

iPaulPro:
Since the update profile transaction allows the caller to change the Founder Reward percentage, it can be argued that this is a financial transaction. Furthermore, changing a username could be considered one of the most sensitive operations on a platform built around reputation.


maebeam:
This is a great change! Thanks for making it.

PR #66, opened 2021-09-15
Buy clout with eth

maebeam:


src/app/crypto.service.ts
```diff
@@ -172,4 +173,38 @@ export class CryptoService {
 
     return bs58check.encode(prefixAndKey);
   }
+
+  // NOTE: Our ETH address uses the bitcion/bitclout derivation path, not the ETH path.
```
redpartyhat, line 177: nit: bitcoin*

src/app/crypto.service.ts
```diff
@@ -172,4 +173,38 @@ export class CryptoService {
...
+  //
+  // Reference implementation: https://github.com/ethereumjs/ethereumjs-util/blob/master/src/account.ts#L249
+  keychainToEthAddress(keychain: HDNode, network: Network): string {
+    // Get the uncompressed key
+    const ec = new EC('secp256k1');
+    const publicKeyEC = ec.keyFromPublic(keychain.publicKey, 'array');
+
+    // ETH uses the last 40 characters of the 64 byte SHA3 Keccak 256
+    const uncompressedKey = Buffer.from(publicKeyEC.getPublic(false, 'array').slice(1));
+    const ethAddress = new Keccak(256).update(uncompressedKey).digest('hex').slice(24);
```
redpartyhat, line 194: In the reference implementation you've linked to in the comment for this function, it says:
```
  // Only take the lower 160bits of the hash
  return keccak(pubKey).slice(-20)
```
^This code only takes 20 bytes while yours takes 40.  Am I reading that correctly? Can you explain what the difference is?

src/app/crypto.service.ts
```diff
@@ -172,4 +173,38 @@ export class CryptoService {
...
+  // NOTE: Our ETH address uses the bitcion/bitclout derivation path, not the ETH path.
+  // This is ugly but we only do it because they're just deposit addresses and we couldn't
+  // backfill this data for existing users because we store the derived private key.
+  // A user can still easily import their account to an ETH client to recover any ETH
+  // by coping the seedHex from local storage.
+  //
+  // We aren't using ethereumjs-util to minimize bundle size and protect from vulnerabilities
+  // until LavaMoat is ready
+  //
+  // Reference implementation: https://github.com/ethereumjs/ethereumjs-util/blob/master/src/account.ts#L249
```
redpartyhat, line 186: Thanks for linking to this reference implementation.  Super helpful!

src/app/crypto.service.ts
```diff
@@ -172,4 +173,38 @@ export class CryptoService {
 
     return bs58check.encode(prefixAndKey);
   }
+
+  // NOTE: Our ETH address uses the bitcion/bitclout derivation path, not the ETH path.
+  // This is ugly but we only do it because they're just deposit addresses and we couldn't
+  // backfill this data for existing users because we store the derived private key.
+  // A user can still easily import their account to an ETH client to recover any ETH
+  // by coping the seedHex from local storage.
```
redpartyhat, line 181: Have you confirmed that this works?

redpartyhat, line 181: Have we checked that this works?  Like that the seedhex in identity perfectly maps to the eth address that this function produces?

maebeam: Yep, verified

src/app/crypto.service.ts
maebeam: 40 hex characters is 20 bytes :)

redpartyhat: Great, makes sense!

src/app/crypto.service.ts
redpartyhat: Okay dope.  I feel pretty good about the addresses then :)

src/app/app.component.ts
```diff
@@ -54,6 +52,9 @@ export class AppComponent implements OnInit {
       this.globalVars.hideGoogle = true;
     }
 
+    // Migrate all accounts
+    this.accountService.migrate();
```
redpartyhat, line 56: Migrate seems dope, love it.

src/app/account.service.ts
```diff
@@ -71,26 +73,31 @@ export class AccountService {
...
-
-    localStorage.setItem(AccountService.usersStorageKey, JSON.stringify(privateUsers));
-
-    return publicKey;
+  // Public Modifiers
+
+  addUser(keychain: HDKey, mnemonic: string, extraText: string, network: Network, google?: boolean): string {
+    const seedHex = this.cryptoService.keychainToSeedHex(keychain);
+    const btcDepositAddress = this.cryptoService.keychainToBtcAddress(keychain, network);
+    const ethDepositAddress = this.cryptoService.keychainToEthAddress(keychain);
```
diamondhands0, line 81: It's fine, but curious: Is there a reason we moved the pubkey generation for BTC/ETH to identity from backend where it was before?

src/app/crypto.service.ts
diamondhands0: +1

PR #71, opened 2021-09-22
Shared secret endpoint

AeonSw4n:
WIP, depends on #53 and derived key backend PR.

Path name is `/get-shared-secrets` and you need to pass following URL params:
- `ownerPublicKey` - base58check of owner
- `derivedPublicKey` - base58check of derived key
- `JWT` - signed by derived key (PR also adds `derivedJwt` to `/derive` endpoint's payload for simplicity)
- `messagePublicKeys` - CSV format, no quotes of public keys in base58check
- `callback`


src/app/account.service.ts
```diff
@@ -74,6 +83,47 @@ export class AccountService {
...
+    const keyEncoder = new KeyEncoder('secp256k1');
+    const encodedPrivateKey = keyEncoder.encodePrivate(privateUser.seedHex, 'raw', 'pem');
+    const jwt = jsonwebtoken.sign({ }, encodedPrivateKey, { algorithm: 'ES256', expiresIn: '30 days' });
+
+    // We compute a derived key JWT with a month-long expiration. This is needed for shared secret endpoint.
+    const encodedDerivedPrivateKey = keyEncoder.encodePrivate(derivedSeedHex, 'raw', 'pem');
+    const derivedJwt = jsonwebtoken.sign({ }, encodedDerivedPrivateKey, { algorithm: 'ES256', expiresIn: '30 days' });
+
+    return {
+      btcDepositAddress: privateUser.btcDepositAddress,
```
maebeam, line 115: Won't they need a new, derived `btcDepositAddress` and `ethDepositAddress` so they can actually sign these transactions with their derived key?

src/app/account.service.ts
```diff
@@ -74,6 +83,47 @@ export class AccountService {
...
+    const derivedPrivateKey = this.cryptoService.seedHexToPrivateKey(derivedSeedHex);
+    const derivedPublicKey = this.cryptoService.privateKeyToDeSoPublicKey(derivedPrivateKey, privateUser.network);
+
+    // By default we authorize this derived key for 10,000 blocks.
+    const expirationBlock = blockHeight + 10000;
+
+    const expirationBlockBuffer = uint64ToBufBigEndian(expirationBlock);
+    const derivedPublicKeyBuffer = derivedPrivateKey.getPublic().encode('array', true);
+    const accessHash = sha256.x2([...derivedPublicKeyBuffer, ...expirationBlockBuffer]);
+    const accessSignature = this.signingService.signBurn(privateUser.seedHex, [accessHash])[0];
```
maebeam, line 103: If this method is being used to sign things other than burns then we should rename it to something like `signHashes`

src/app/account.service.ts
```diff
@@ -139,6 +189,20 @@ export class AccountService {
     this.setPrivateUsersRaw(privateUsers);
   }
 
+  getSharedSecret(ownerPublicKey: string, publicKey: string): string {
+    const privateUsers = this.getPrivateUsers();
+    if (ownerPublicKey in privateUsers) {
```
maebeam, line 194: Always return early instead of nesting if statements. `If not in, return ''`

src/app/app.component.ts
```diff
@@ -69,20 +80,28 @@ export class AppComponent implements OnInit {
...
 
-    if (this.globalVars.webview || this.globalVars.inTab || this.globalVars.inFrame()) {
-      // We must be running in a webview OR opened with window.open OR in an iframe to initialize
-      this.identityService.initialize().subscribe(res => {
-        this.globalVars.hostname = res.hostname;
-        if (this.globalVars.isFullAccessHostname()) {
-          this.globalVars.accessLevelRequest = AccessLevel.Full;
-        }
 
+    if (this.globalVars.webview || this.globalVars.inTab || this.globalVars.inFrame()) {
```
maebeam, line 89: Which one of these is `true` when there's a callback and why? Seems like none of them should be

src/app/backend-api.service.ts
```diff
@@ -83,4 +122,30 @@ export class BackendAPIService {
       ReferralHash,
     });
   }
+
+  GetUserDerivedKeys(
+    ownerPublicKey: string
+  ): Observable< { [key: string]: DerivedKey } > {
+    return new Observable<{ [key: string]: DerivedKey }>(subscriber => {
+      this.httpClient.post<any>(
+        `${this.endpoint}/get-user-derived-keys`,
```
maebeam, line 131: You need to update the `Caddyfile` to allow access to this endpoint

src/app/backend-api.service.ts
```diff
@@ -31,6 +44,32 @@ export class BackendAPIService {
...
+          userProfiles[publicKey] = {};
+        }
+        this.GetUsersStateless(publicKeys).subscribe(res => {
+          for (const user of res.UserList) {
+            userProfiles[user.PublicKeyBase58Check] = {
+              username: user.ProfileEntryResponse?.Username,
+              profilePic: user.ProfileEntryResponse?.ProfilePic,
+            };
+          }
+          subscriber.next(userProfiles);
```
maebeam, line 65: While this implementation is fine, it's not ideal. If you want to try your hand at RXJS operators (which can be annoying but very powerful), checkout `backend-api.ts` in the frontend repo. The best way to implement this type of behavior is to do something akin to:
```
const userProfiles: {[key: string]: any} = {};
const req = this.GetUsersStateless(publicKeys);
if () {
  req.pipe(
    map((res) => {
      // ...
      return userProfiles;
    })
  )
} else {
  return of(userProfiles);
}

```

src/app/backend-api.service.ts
```diff
@@ -83,4 +122,30 @@ export class BackendAPIService {
...
+          if (res.DerivedKeys.hasOwnProperty(derivedKey)) {
+            derivedKeys[res.DerivedKeys[derivedKey]?.DerivedPublicKeyBase58Check] = {
+              derivedPublicKeyBase58Check: res.DerivedKeys[derivedKey]?.DerivedPublicKeyBase58Check,
+              ownerPublicKeyBase58Check: res.DerivedKeys[derivedKey]?.OwnerPublicKeyBase58Check,
+              expirationBlock: res.DerivedKeys[derivedKey]?.ExpirationBlock,
+              isValid: res.DerivedKeys[derivedKey]?.IsValid,
+            };
+          }
+        }
+        subscriber.next(derivedKeys);
```
maebeam, line 147: Same comment here as above regarding RXJS operators

src/app/derive/derive.component.html
```diff
@@ -0,0 +1,52 @@
+<app-banner></app-banner>
+
+<div class="container home-container" *ngIf="globalVars.inTab || globalVars.webview">
```
maebeam, line 3: Again I'm confused about why `webview` or `inTab` would be true here

src/app/derive/derive.component.html
```diff
@@ -0,0 +1,52 @@
...
+  <div class="font-weight-bold fs-30px mb-20px" *ngIf="!withCallback">
+    Grant signing access to {{ globalVars.hostname }}
+  </div>
+  <div class="font-weight-bold fs-30px mb-20px" *ngIf="withCallback">
+    Grant signing access
```
maebeam, line 8: I think we do need to add an `?app=Flick App` param

src/app/derive/derive.component.html
```diff
@@ -0,0 +1,52 @@
+<app-banner></app-banner>
+
+<div class="container home-container" *ngIf="globalVars.inTab || globalVars.webview">
+  <div class="font-weight-bold fs-30px mb-20px" *ngIf="!withCallback">
```
maebeam, line 4: When would you ever not use a callback?

src/app/derive/derive.component.html
```diff
@@ -0,0 +1,52 @@
...
+          This permission can be revoked anytime and it will automatically expire in about 30 days.
+      </div>
+    </div>
+    <div *ngIf="withCallback">
+      <div class="fs-18px mt-30px">
+        <span class="fs-underline"><b>You're about to grant signing access:</b></span><br/>
+        This means the mobile application will be able to sign transactions on your behalf as if it knew your seed.
+        This permission can be revoked anytime and it will automatically expire in about 30 days.
+        <br/><br/>
+        <span class="fs-underline"><b>DISCLAIMER:</b></span><br/>
```
maebeam, line 46: I think we could remove this but I'm not sure. Will defer to diamond hands

src/app/identity.service.ts
```diff
@@ -67,6 +72,24 @@ export class IdentityService {
...
+      let paramString = '?';
+      for (const key in payload.derivedUserInfo) {
+        if (payload.derivedUserInfo.hasOwnProperty(key)) {
+          paramString += key + '=' + (payload.derivedUserInfo as any)[key].toString() + '&';
+        }
+      }
+      paramString = paramString.slice(0, -1);
+      window.location.href = this.globalVars.callback.href + paramString;
+    } else {
+      this.cast('derive', payload);
```
maebeam, line 89: Why would this ever be used in a desktop browser context?

src/app/log-in/log-in.component.ts
```diff
@@ -33,52 +33,19 @@ export class LogInComponent implements OnInit {
...
+        this.allUsers = profiles;
       });
-    }
+
+    // Set showAccessLevels
+    this.showAccessLevels = !this.globalVars.isFullAccessHostname();
   }
 
   launchGoogle(): void {
-    const redirectUri = new URL(`${window.location.origin}/${RouteNames.AUTH_GOOGLE}`);
```
maebeam, line 62: Thanks for not duplicating code!

src/app/shared-secret/shared-secret.component.ts
```diff
@@ -0,0 +1,115 @@
...
+            for (const messageKey of this.messagePublicKeys){
+              const sharedSecret = this.accountService.getSharedSecret(this.ownerPublicKeyBase58Check, messageKey);
+              if (sharedSecret === ''){
+                this.errorMsg = 'Failed to calculate shared secrets';
+                return;
+              }
+              sharedSecrets.push(sharedSecret);
+            }
+
+            let paramString = '?sharedSecrets=';
```
maebeam, line 106: Please use `HTTPParams`:

```
    let httpParams = new HttpParams();
    if (this.isTestnet) {
      httpParams = httpParams.append("testnet", "true");
    }

    if (params?.public_key) {
      httpParams = httpParams.append("public_key", params.public_key);
    }

    const paramsStr = httpParams.toString();
    if (paramsStr) {
      url += `?${paramsStr}`;
    }
```

src/app/app.component.html
```diff
@@ -1,3 +1,8 @@
 <div *ngIf="!loading" class="full-screen">
-  <router-outlet></router-outlet>
+  <div *ngIf="isCallbackValid">
+    <router-outlet></router-outlet>
+  </div>
+  <div *ngIf="!isCallbackValid">
```
maebeam, line 5: I don't like polluting the app component with error handling like this. Can we not worry about this here and just display an error in any component that requires a callback?

src/app/account.service.ts
AeonSw4n: We need to consider how we treat btc & eth deposit addresses for derived keys in core/backend.

src/app/app.component.ts
AeonSw4n: Didn't really have a way to verify. Seems like one of them does since it works for Andy, but I've reorganized the conditions just to make sure.

andyboyd: I was also passing `webview=true` to make it work. Would be best if we didn't need to do that, though.

src/app/backend-api.service.ts
AeonSw4n: ugh, RXJS... great comment though, thanks

src/app/shared-secret/shared-secret.component.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor(
+    private accountService: AccountService,
+    private backendApi: BackendAPIService,
+    private cryptoService: CryptoService,
+    public globalVars: GlobalVarsService,
+  ) { }
+
+  ngOnInit(): void {
+    if (!this.globalVars.callback){
```
AeonSw4n, line 30: +isCallbackValid

AeonSw4n (commented):
in-class review

src/app/app.component.ts
AeonSw4n: @andyboyd, I fixed it now. Also added `?app` param we've talked about.

src/app/global-vars.service.ts
```diff
@@ -18,11 +18,14 @@ export class GlobalVarsService {
   hideGoogle = false;
   hideJumio = false;
 
+  callback: URL | null = null;
+  isCallbackValid = true;
```
maebeam, line 22: Let's default this to false and set it to true when we successfully parse the callback

maebeam (approved):
Ready for final review from @diamondhands0!

src/app/account.service.ts
```diff
@@ -74,6 +83,54 @@ export class AccountService {
     }
   }
 
+  getDerivedUser(publicKey: string, blockHeight: number): DerivedUserInfo{
```
diamondhands0, line 86: nit: Rename to getDerivedPrivateUser. Also rename DerivedUserInfo to DerivedPrivateUserInfo

src/app/account.service.ts
```diff
@@ -139,6 +196,20 @@ export class AccountService {
     this.setPrivateUsersRaw(privateUsers);
   }
 
+  getSharedSecret(ownerPublicKey: string, publicKey: string): string {
```
diamondhands0, line 199: nit: Rename to getPrivateSharedSecret

src/lib/bindata/util.ts
```diff
@@ -28,3 +28,20 @@ export const uvarint64ToBuf = (uint: number): Buffer => {
 
   return new Buffer(result);
 };
+
+export const uint64ToBufBigEndian = (uint: number): Buffer => {
```
AeonSw4n, line 32: DH: Where'd you get this from. Are you 100% sure it works?

src/app/derive/derive.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    // Set withCallback
+    this.withCallback = this.globalVars.callback !== null;
+  }
+
+  launchGoogle(): void {
+    this.googleDrive.launchGoogle();
+  }
+
+  selectAccount(publicKey: string): void {
```
AeonSw4n, line 48: DH: nit: I don't like that this has the same name as the log-in component. Can we rename this to selectAccountAndDeriveKey?

src/app/app.component.ts
```diff
@@ -51,12 +51,26 @@ export class AppComponent implements OnInit {
     if (params.get('hideGoogle')) {
       this.globalVars.hideGoogle = true;
     }
-    
+
+    // Callback should only be used in mobile applications, where payload is passed through URL parameters.
+    if (params.get('callback')) {
+      try {
+        this.globalVars.callback = new URL(params.get('callback') as string);
+        this.globalVars.isCallbackValid = true;
```
AeonSw4n, line 59: How is callback related to hostname? Do we want to enforce that the callback has the same hostname as the thing that called this? I don't think it's necessary but put some juices on it and lmk.

src/types/identity.ts
```diff
@@ -28,6 +28,19 @@ export interface PublicUserInfo {
   version: PrivateUserVersion;
 }
 
+export interface DerivedUserInfo {
```
AeonSw4n, line 31: DH: nit: Change this to DerivedPrivateUserInfo since this contains key material.

src/app/backend-api.service.ts
```diff
@@ -83,4 +120,31 @@ export class BackendAPIService {
...
+    ownerPublicKey: string
+  ): Observable< { [key: string]: DerivedKey } > {
+    const derivedKeys: { [key: string]: DerivedKey } = {};
+    const req = this.httpClient.post<any>(
+      `${this.endpoint}/get-user-derived-keys`,
+      {
+        PublicKeyBase58Check: ownerPublicKey,
+      },
+    );
+    return req.pipe(
```
diamondhands0, line 134: How does the backend know what your derived keys are? Is there a backend PR I'm missing here?

src/app/identity.service.ts
```diff
@@ -67,6 +73,23 @@ export class IdentityService {
...
+  }): void {
+    if (this.globalVars.callback) {
+      // If callback is passed, we redirect to it with payload as URL parameters.
+      let httpParams = new HttpParams();
+      for (const key in payload.derivedUserInfo) {
+        if (payload.derivedUserInfo.hasOwnProperty(key)) {
+          httpParams = httpParams.append(key, (payload.derivedUserInfo as any)[key].toString());
+        }
+      }
+      window.location.href = this.globalVars.callback.href + `?${httpParams.toString()}`;
```
diamondhands0, line 87: Ah, so much less clever. Good job :)

src/app/shared-secret/shared-secret.component.ts
```diff
@@ -0,0 +1,116 @@
...
+      this.errorMsg = 'Message Public Keys required';
+      console.error(this.errorMsg);
+      return;
+    }
+
+    this.backendApi.GetUserDerivedKeys(this.ownerPublicKeyBase58Check).subscribe( keys => {
+      if (this.derivedPublicKeyBase58Check in keys) {
+        this.backendApi.GetAppState().subscribe( state => {
+          const blockHeight = state.BlockHeight;
+          const key = keys[this.derivedPublicKeyBase58Check];
```
diamondhands0, line 95: Could key be null if the derived key isn't in keys?

AeonSw4n: I think it would fail the check at line #92, or not?

diamondhands0: I think you're right but can you just Google it and double check? If we get an error here it could cause issues.

src/lib/bindata/util.ts
AeonSw4n: @diamondhands0 this is essentially doing the same thing as [`PutUint64`](https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/encoding/binary/binary.go;l=130) from the binary package that we use in core in [`EncodeUint64`](https://github.com/deso-protocol/core/blob/be25e74a93c93068cc4314c65fb26f99ddc70e52/lib/db_utils.go#L1838)  - and then we use this in [`_verifyAccessSignature()`](https://github.com/deso-protocol/core/blob/be25e74a93c93068cc4314c65fb26f99ddc70e52/lib/block_view.go#L7939 ). I wrote this myself, tested it, and it seems to work. Regardless, if it fails for some reason, this would only impact the UX. Core will reject the discrepancy in access signature so user will have to generate another derived key.

src/app/app.component.ts
AeonSw4n: @diamondhands0 I believe mobile clients have localhost hostname. To make callbacks safer, we had this idea to whitelist URL schemes for callbacks on a case by case basis, starting with e.g. `identity://`. I remember it's possible to have a custom, static scheme on iOS, but not sure if that's the case for Android. We should discuss more on Slack.

src/app/backend-api.service.ts
AeonSw4n: Resolved, right?

src/app/shared-secret/shared-secret.component.ts
AeonSw4n: I'm ready to bet that we will never get to this point and have an undefined/null key; but JUST in case I added an early return.

src/app/app.component.ts
diamondhands0: What's safer about enforcing identity://? Seems equally safe/unsafe compared to allowing an arbitrary URL scheme doesn't it? Either way, I think this is a fine threat model as-is. Basically, if the user hits selectAccountAndDeriveKey, then they're at the mercy of whatever callback URL the caller set, which makes sense.

Resolving.

src/app/app.component.ts
```diff
@@ -51,12 +51,26 @@ export class AppComponent implements OnInit {
...
+      }
+    }
+
     if (params.get('hideJumio')) {
       this.globalVars.hideJumio = true;
     }
-    
-    const referralCode = params.get('referralCode')
+
+    if (params.get('app')) {
```
diamondhands0, line 69: Oh yeah I forgot to ask.. what's the point of this app param?

diamondhands0 (approved):
Excellent work Piotr! This change seems fairly safe. Just get rid of the app param thing like we talked about, and @maebeam give it a good test with the UI and everything before full deploy.

PR #81, opened 2021-10-14
remove full access from 3rd party apps

tijno:
This change was a major breach of the CIP process in my view and should have never been implemented. I propose to remove full access permissions from 3rd party apps with immediate effect.

### DIFF
--- a/src/environments/environment.prod.ts
+++ b/src/environments/environment.prod.ts
@@ -2,7 +2,7 @@ export const environment = {
   production: true,
   hostname: 'identity.deso.org',
   nodeHostname: 'bitclout.com',
-  fullAccessHostnames: ['bitclout.com', 'bitclout.green', 'bitclout.blue', 'localhost', 'buy.deso.org', 'diamondapp.com'],
+  fullAccessHostnames: ['bitclout.com', 'bitclout.green', 'bitclout.blue', 'localhost', 'buy.deso.org'],
   noAccessHostnames: [''],
   jumioSupported: true,
 };


tijno:
Yeah and personally as Diamond is a separate for profit org - it should not have same whitelisting permissions as bitclout.com.

Hence my suggestion to remove it from the full access whitelist.

diamondhands0:
Hey all, thank you for pointing this out; whitelisting diamondapp.com allows it to see all accounts, not just ones you logged into, but this is temporary. We whitelisted diamondapp.com as a hotfix to allow Diamond to launch while we fix a bug in Identity whereby it was showing a popup even after granting level 4 access. Once this bug is fixed, diamondapp.com will be removed from the whitelist, and it will only have access to the accounts you explicitly grant it access to just like all third-party apps.

The launch of Diamond gave the core dev team a great opportunity to optimize the process of launching one's own independent third-party node, and we appreciate your patience with us as we iron out the last few remaining kinks over the next week or two.

tijno:
ive submitted the PR 
https://github.com/deso-protocol/frontend/pull/486

But I see that diamond App did already implement this 4 days ago. And that it it present.

So this should solve the level 4 issue already.

tijno:
diamondapp is still listed as a full access node name - is that really required? If so - please explain remaining reasons and why they dont apply to other 3rd party nodes.

tijno:
@diamondhands0 diamondapp is still listed.

you mentioned this was temporary in your reply above 4 months ago.

maybe its time to remove it ?

cc @superzordon

tijno:
closing this

AdityaChaudhary0005:
it was fun till it lasted @tijno ...

PR #99, opened 2021-11-27
Fix logout

AeonSw4n:


diamondhands0 (approved):
Interesting solution. Seems like it should work. Good job.

maebeam:
Can you add a comment in logout component?

AeonSw4n:
@maebeam added the comment

maebeam (approved):

PR #101, opened 2021-11-30
Add support for approve flow for ETH and BTC purchases of DeSo

lazynina:


maebeam (approved):
Can you review as well @AeonSw4n

diamondhands0 (approved):

src/app/approve/approve.component.ts
```diff
@@ -58,6 +61,19 @@ export class ApproveComponent implements OnInit {
...
+        this.transactionDescription = 'an ethereum transaction to purchase DeSo';
+        this.publicKey = params.publicKey;
+        this.ethTx = params.ethTx;
+        return;
+      }
+      if (params.btcTx) {
+        this.transactionDescription = 'a btc transaction to purchase DeSo';
+        this.publicKey = params.publicKey;
+        this.btcTx = params.btcTx;
+        return;
```
AeonSw4n, line 74: Hmmm I'm not sure I like this, couldn't you potentially pass a DeSo transaction in `ethTx` or `btcTx` and pretend it's an eth or btc transaction but in reality, it's a basic transfer sending all your DESO to another public key? Seems like that would be possible considering we use the publicKey from URL param to fetch seedHex to sign the txn. Wyt?

maebeam: This is actually a good point, someone could abuse this to sign a hash for a deso transaction. I'm not sure there's anything we can do about this though

AeonSw4n: Yeah idk a good solution either. Seems like adding serialization and better description could fix this, but that's likely overdoing it.

lazynina: It looks like we currently do some slicing and dicing in signTransaction that we don't in signHashes, so I think it'll be okay. The signETH wouldn't work for DeSo transactions I don't think, but I'm not positive someone couldn't take the signature and turn it into something valid. 

I'll see if there's a way to test if a transaction is indeed a BTC or an ETH transaction.

lazynina: Or I could try to parse the hex as a DeSo transaction and if it appears to be a DeSo transaction, we treat it as such. So basically we test that a transaction is not DeSo before asking the user to sign it. deserializing is the ideal option though.

AeonSw4n: Yeah that could work

PR #115, opened 2022-01-25
Phone verification mobile cb

imekinox:
Since we can't use web view on mobile to sign in users on mobile, we are using Chrome Custom Tabs for Android & SafariServices/AuthenticationServices for iOS. Given the security restrictions for those, we can't use postMessage to communicate with the identity service. 

In this PR I'm adding the possibility to call a callback instead of postMessage in case it's provided in the URL. 
Also, I'm changing ( if callback is provided ) the "Continue to localhost" message at the end of the phone verification process to just "Continue" so it's less confusing for the user.

### DIFF
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -77,7 +77,18 @@ export class IdentityService {
     jumioSuccess?: boolean,
     phoneNumberSuccess?: boolean,
   }): void {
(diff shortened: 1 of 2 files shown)


lazynina (approved):
@AeonSw4n for second review, but this looks good to me.

Issue #109, opened 2021-12-30
Incorrect spending formatting in Approve


AeonSw4n:
![image-2](https://user-images.githubusercontent.com/29924594/147764388-55452afd-d3e3-45e2-a56a-95f0845c9b4f.png)
![image](https://user-images.githubusercontent.com/29924594/147764390-29476211-e0f7-4d90-8f91-d3ffaba5e1ca.png)

AeonSw4n:
Resolved in #133!

PR #118, opened 2022-02-15
Signing arbitrary messages

iPaulPro:
Adds the ability to sign arbitrary plain text messages with Identity.

These changes include updating the `sign` iframe endpoint to support a `messageHex` parameter, and the Approve component to support signing messages.

MetaMask's [personal_sign](https://github.com/ethereum/go-ethereum/pull/2940) method was referenced when determining the structure for the signed message response.

`sign` returns an ECDSA signature for sha256.x2(message). A prefix is added to the message before hashing to prevent unwanted signing of a transaction. The hash is calculated with:

```
sha256.x2("DeSo Signed Message:\n"${message.length}${message})
```

A `signedMessageHex` is returned as part of the message payload when successful.

Here is the relevant snippet:

```ts
sign(seedHex: string, messageHex: string): string {
  const privateKey = this.cryptoService.seedHexToPrivateKey(seedHex);

  const contentBytes = new Buffer(messageHex, 'hex');
  const contentLength = uvarint64ToBuf(contentBytes.length);

  // We add a prefix to prevent the caller from signing a transaction
  const prefixBytes = new Buffer(this.globalVars.signPrefix, 'utf-8');

  const messageBytes = Buffer.concat([
    prefixBytes,
    contentLength,
    contentBytes,
  ]);
  const messageBytesLength = uvarint64ToBuf(messageBytes.length);
  const messageHash = new Buffer(sha256.x2(messageBytes), 'hex');

  const signature = privateKey.sign(messageHash);
  const signatureBytes = new Buffer(signature.toDER());
  const signatureLength = uvarint64ToBuf(signatureBytes.length);

  const signedMessageBytes = Buffer.concat([
    messageBytesLength,
    messageBytes,
    signatureLength,
    signatureBytes
  ])

  return signedMessageBytes.toString('hex');
}
```

An example of an iframe message:

```js
const payload = {
  accessLevel: user.accessLevel,
  accessLevelHmac: user.accessLevelHmac,
  encryptedSeedHex: user.encryptedSeedHex,
  messageHex: Buffer.from(text, 'hex')
}

const msg = {
  id: id,
  service: 'identity',
  method: 'sign',
  payload: payload
}

iframe.contentWindow.postMessage(msg, "*")
```

And the approval window URL construction, which requires that you also pass the `PublicKeyBase58Check` as `publicKey`:

```js
const url = new URL('https://identity.deso.org/approve')
const messageHex = Buffer.from(text, 'hex')
url.searchParams.set('messageHex', messageHex)
url.searchParams.set('publicKey', publicKey)
```

Here's a basic example of signature verification:

```js
const EC = require('elliptic').ec
const bs58check = require('bs58check')
const sha256 = require('sha256')

const verifySignedMessage = (publicKeyBase58Check, signedMessageHex) => {
  const signedMessageBytes = Buffer.from(signedMessageHex, 'hex')

  const messageBytesLength = signedMessageBytes[0]
  const messageBytes = signedMessageBytes.slice(1, messageBytesLength + 1)

  const messageHash = Buffer.from(sha256.x2(messageBytes), 'hex')
  const signature = signedMessageBytes.slice(messageBytesLength + 2).toString('hex')

  const decoded = bs58check.decode(publicKeyBase58Check)
  const publicKey = Uint8Array.from(decoded).slice(3)
  const ec = new EC('secp256k1')
  const publicKeyEC = ec.keyFromPublic(publicKey, 'array')

  return publicKeyEC.verify(messageHash, signature)
}
```

I will submit a PR to update the docs after this is merged. I also plan to add a Util class to [deso.js](https://github.com/bitclouthunt/deso.js/) with some functions to support this flow.

### DIFF
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -42,12 +42,13 @@ import bs58check from 'bs58check';
   styleUrls: ['./approve.component.scss']
 })
 export class ApproveComponent implements OnInit {
-  transaction: any;
+  transaction?: any;
   publicKey: any;
-  transactionHex: any;
+  transactionHex?: any;
   username: any;
   transactionDescription: any;
   transactionDeSoSpent: string | boolean = false;
+  messageHex?: string;
 
   constructor(
     private activatedRoute: ActivatedRoute,
@@ -61,15 +62,28 @@ export class ApproveComponent implements OnInit {
 
   ngOnInit(): void {
     this.activatedRoute.queryParams.subscribe(params => {
-      this.transactionHex = params.tx;
-      this.backendApi.GetTransactionSpending(this.transactionHex).subscribe( res => {
-        this.transactionDeSoSpent = res ? this.nanosToUnitString(res) : false;
-      });
-      const txBytes = new Buffer(this.transactionHex, 'hex');
-      this.transaction = Transaction.fromBytes(txBytes)[0];
-      this.publicKey = this.base58KeyCheck(this.transaction.publicKey);
+      if (params.messageHex) {
+        this.messageHex = params.messageHex;
+        this.publicKey = params.publicKey;
 
-      this.generateTransactionDescription();
+        // Ensure the public key passed is one we know
+        const encryptedUser = this.accountService.getPublicKeys().includes(this.publicKey)
+        if (!encryptedUser) {
+          this.finishFlow();
+        }
+
+        this.transactionDescription = 'sign a message:';
+      } else {
+        this.transactionHex = params.tx;
+        this.backendApi.GetTransactionSpending(this.transactionHex).subscribe(res => {
+          this.transactionDeSoSpent = res ? this.nanosToUnitString(res) : false;
+        });
+        const txBytes = new Buffer(this.transactionHex, 'hex');
+        this.transaction = Transaction.fromBytes(txBytes)[0];
+        this.publicKey = this.base58KeyCheck(this.transaction.publicKey);
+
+        this.generateTransactionDescription();
+      }
     });
   }
 
@@ -78,14 +92,22 @@ export class ApproveComponent implements OnInit {
   }
 
   onSubmit(): void {
-    const signedTransactionHex = this.signingService.signTransaction(this.seedHex(), this.transactionHex);
-    this.finishFlow(signedTransactionHex);
+    if (this.messageHex) {
+      const signedMessageHex = this.signingService.sign(this.seedHex(), this.messageHex);
+      this.finishFlow({signedMessageHex});
+    } else {
+      const signedTransactionHex = this.signingService.signTransaction(this.seedHex(), this.transactionHex);
+      this.finishFlow({signedTransactionHex});
+    }
   }
 
-  finishFlow(signedTransactionHex?: string): void {
+  finishFlow(payload: {
+    signedTransactionHex?: string,
+    signedMessageHex?: string
+  } = {}): void {
     this.identityService.login({
       users: this.accountService.getEncryptedUsers(),
-      signedTransactionHex,
+      ...payload
     });
   }
 
@@ -302,5 +324,17 @@ export class ApproveComponent implements OnInit {
       }
       return outputString;
     })));
+  }
+
+  getMessageAsText() {
+    if (!this.messageHex) {
+      return null;
+    }
+    try {
+      const buff = Buffer.from(this.messageHex, 'hex');
+      return buff.length === 32 ? this.messageHex : buff.toString('utf8');
+    } catch (e) {
+      return this.messageHex;
+    }
   }
 }
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -62,6 +62,10 @@ export class ApproveComponent implements OnInit {
 
   ngOnInit(): void {
     this.activatedRoute.queryParams.subscribe(params => {
+      if (params.tx && params.messageHex) {
+        console.error('Found transactionHex and messageHex in query parameters. Only messageHex will be used.');
+      }
+
       if (params.messageHex) {
         this.messageHex = params.messageHex;
         this.publicKey = params.publicKey;
--- a/src/app/identity.service.ts
+++ b/src/app/identity.service.ts
@@ -146,25 +146,23 @@ export class IdentityService {
   private handleSign(data: any): void {
     const { id, payload: { encryptedSeedHex, transactionHex, messageHex } } = data;
 
-    const seedHex = this.cryptoService.decryptSeedHex(encryptedSeedHex, this.globalVars.hostname);
-
     if (transactionHex && messageHex) {
       console.error('Found transactionHex and messageHex in payload. Only messageHex will be used.');
     }
 
     if (messageHex) {
       // No transaction or spending here, so we just check if approval is required.
-      if (!this.hasAccessLevel(data, AccessLevel.ApproveLarge)) {
-        this.respond(data.id, {approvalRequired: true});
+      if (!this.approve(data, AccessLevel.ApproveLarge)) {
         return;
       }
 
+      const seedHex = this.cryptoService.decryptSeedHex(encryptedSeedHex, this.globalVars.hostname);
       const signedMessageHex = this.signingService.sign(seedHex, messageHex);
 
       this.respond(id, {
         signedMessageHex,
       });
-    } else if (transactionHex) {
+    } else {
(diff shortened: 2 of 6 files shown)


src/app/approve/approve.component.ts
```diff
@@ -78,14 +92,22 @@ export class ApproveComponent implements OnInit {
   }
 
   onSubmit(): void {
-    const signedTransactionHex = this.signingService.signTransaction(this.seedHex(), this.transactionHex);
-    this.finishFlow(signedTransactionHex);
+    if (this.messageHex) {
```
HPaulson, line 95:
```suggestion
    if (this.messageHex && this.transactionHex) {} // Some error
    if (this.messageHex) {
```

src/app/identity.service.ts
```diff
@@ -143,34 +144,49 @@ export class IdentityService {
...
-    const requiredAccessLevel = this.getRequiredAccessLevel(transactionHex);
+    const seedHex = this.cryptoService.decryptSeedHex(encryptedSeedHex, this.globalVars.hostname);
 
-    // In the case that approve() fails, it responds with a message indicating
-    // that approvalRequired = true, which the caller can then uses to trigger
-    // the /approve UI.
-    if (!this.approve(data, requiredAccessLevel)) {
-      return;
-    }
+    if (transactionHex) {
```
HPaulson, line 151:
```suggestion
    if (transactionHex && messageHex ) {} // Some error
    if (transactionHex && !messageHex) {
```
Here we check if there's no messageHex too in accordance with destructive priority

HPaulson (changes requested):
When signing via this method, messages & tx's should be mutually exclusive. There will never be a case where someone should intentionally pass both a message & tx, so if they do this, it's a good idea to error. While the approach of prioritizing one over the other (Right now, this prioritizes a Message in `approve.component`, but a transaction in `identity.service`) does work, it is still prone to the accidental case where someone passes both. This could cause a situation where an engineering mistake causes the signature of an unexpected message/tx (i.e signing a message when they meant to sign a tx)

Some nuances:
- Right now, Identity doesn't have solid error system. This is, of course, out of the scope of this PR, but it's worth mentioning here. Since this doesn't exist, and isn't related to this PR, I'm not exactly sure how you'd effectively "reject" or throw an error when both message & tx are passed (Which is why I left the error as a comment). This item should probably be addressed by a maintainer 
- If, because of the above, it doesn't make sense to error for both, then the message should be the priority. Right now this is the case for user approvals, but not for background signing. The reason I believe this is because I'd consider signing a message the "less" destructive option compared to signing an actual transaction. Of course, though, not allowing both is ideal; this should be considered a "backup" if mutual exclusivity isn't possible due to lack of an erroring system.

src/app/approve/approve.component.ts
```diff
@@ -61,15 +62,28 @@ export class ApproveComponent implements OnInit {
...
-      this.transactionHex = params.tx;
-      this.backendApi.GetTransactionSpending(this.transactionHex).subscribe( res => {
-        this.transactionDeSoSpent = res ? this.nanosToUnitString(res) : false;
-      });
-      const txBytes = new Buffer(this.transactionHex, 'hex');
-      this.transaction = Transaction.fromBytes(txBytes)[0];
-      this.publicKey = this.base58KeyCheck(this.transaction.publicKey);
-
-      this.generateTransactionDescription();
+      if (params.messageHex) {
```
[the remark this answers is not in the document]
iPaulPro: Thanks for your review! Addressed in dace80f

iPaulPro:
> This could cause a situation where an engineering mistake causes the signature of an unexpected message/tx (i.e signing a message when they meant to sign a tx)

Thanks for the review! I updated the code to log an error when both are present, and to always prefer `messageHex`. It should be noted, though, that even signing a transaction accidentally is unlikely to be destructive since it still needs to be broadcasted.

src/app/signing.service.ts
```diff
@@ -160,6 +160,38 @@ export class SigningService {
...
 
+  sign(seedHex: string, messageHex: string): string {
+    const privateKey = this.cryptoService.seedHexToPrivateKey(seedHex);
+
+    const contentBytes = new Buffer(messageHex, 'hex');
+    const contentLength = uvarint64ToBuf(contentBytes.length);
+
+    // We add a prefix to prevent the caller from signing a transaction
+    const prefix = 'DeSo Signed Message:\n';
+    const prefixBytes = new Buffer(prefix, 'hex');
```
AeonSw4n, line 171: Ah so I got some bad news. You just got scammed by the Buffer library. You see we're living in a very degenerate world out there on the blockchain and we need to hand-verify all things. But worry not, I catch everything! So I was playing around with this, and turns out that while your algo is good overall, this lil `'hex'` guy is f-ing us. Try to sign something by hand and put `console.log(prefixBytes)` after this statement. Turns out `prefixBytes = [222]`! Not what we want. We would instead expect the `"DeSo Signed Message:\n"` string to turn into something like `[68 101 83 111 32 83 105 103 110 101 100 32 77 101 115 115 97 103 101 58 10]`. To do that, we need to change this into `new Buffer(prefix, 'utf-8')` just to make sure the Buffer dude knows we're passing an utf-8 string instead of a hex string.

AeonSw4n (changes requested):
Loooks great! Mostly nits and one comment about the Buffer library doing some freaky things. Thank you so much for this contribution! Also, I've made a PR to your fork https://github.com/iPaulPro/identity/pull/1 that also fixes #109. Two birds with one stone.

src/app/signing.service.ts
iPaulPro: Fixed in 00aed39!

AeonSw4n (approved):
👍

iPaulPro:
anybody there?

iPaulPro:
Can someone take a minute to press the merge button? Or maybe let me know what the hold up is... @lazynina @diamondhands0 

Maybe I should translate to a language you'll understand: 

DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO?

lazynina:
> Can someone take a minute to press the merge button? Or maybe let me know what the hold up is... @lazynina @diamondhands0
> 
> Maybe I should translate to a language you'll understand:
> 
> DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO?

@iPaulPro - @diamondhands0 needs to review this PR before we can merge it, that is the reason for the hold up. thanks for the contribution!

diamondhands0:
@iPaulPro I don't understand. Did you mean:

DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO?

iPaulPro:
Not sure why I waste my time, but lesson finally learned. Thanks for nothing.

HPaulson:
Embarrassing. @diamondhands0

tijno:
Disappointing. This will
enable such a great project. I get that changes to identity need approval - but that is no reason to not give clear feedback and either decline, request changes or approve.

brootle:
I didn't have time to review this PR from @iPaulPro but I am sure it's great! Will get back to it after the war, have patience!

DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO DAO?

diamondhands0:
Hey guys, I didn't realize this was blocking someone's product; thought it was a "practice PR" or else I would have prioritized it more aggressively.

Some context also: Generally, changes are easy to review and go in quickly-- but Identity is a major exception. With Identity we always require 3 core devs to review it with a fine-toothed comb because exploits in this part of the code are so costly.

That is to say that the dev experience of contributing to identity is much more stringent than contributing to frontend, backend, or even core, which is why it takes a while for us to process it (requires full concentration and discussion with other core devs). You can ask @AeonSw4n who has contributed to all four about how much longer his identity changes took than his core changes.

On top of all that, this is a pretty serious change in that it allows identity to sign arbitrary messages, which could be used to trick users into sending all of their funds somewhere else without a sufficiently-descriptive message.

I'm going to discuss this with @AeonSw4n tomorrow, though, and hopefully we can get comfortable with this. If not, the alternative would be to fork Identity, which is open-source, and self-host the modified iFrame. But this hopefully won't be necessary.

Btw, I was just joking with the DAO reply. I take all PR's seriously, and this one has been on my plate for a while.

brootle:
Actually yes, if somebody wants to use proposed functionality, he/she can fork Identity, merge this PR, and host this updated Identity to use for their apps. I would be worried if some Identity related PR would be approved and merged in like 1-2 days without discussing if it's even necessary for majority of DeSo protocol users, not to mention possible security risks.

PR #119, opened 2022-02-17
Add ability to pass public key to derived key route

superzordon:
This update would allows a public key to be passed to the `derive` endpoint in identity. This could be useful in instances where a server-side application wants to use derived keys, but wants to avoid sending the private key over https. With this update, that application could generate the public/private keypair server-side, and only pass in the public key to generate the accessSignature used for authorizing the derived key.


src/app/derive/derive.component.ts
```diff
@@ -38,6 +40,13 @@ export class DeriveComponent implements OnInit {
 
     // Set derive to true
     this.globalVars.derive = true;
+
+    this.route.queryParams.subscribe((params) => {
+      if (params.derivedPublicKeyBase58Check) {
+        this.derivedPublicKeyBase58Check = params.derivedPublicKeyBase58Check;
```
AeonSw4n, line 46: do we need the local `derivedPublicKeyBase58Check` if we also store it in globalVars?

AeonSw4n (changes requested):
Looks great! Just some minor suggestions & questions.

PR #120, opened 2022-02-19
Update copy to use DeSo Key terminology

triplegreenshell:


iPaulPro:
Why would we call it a "key" when it is, in fact, a seed phrase?

diamondhands0:
@iPaulPro @AdityaChaudhary0005 @tijno, wanted to thank you all for this input and provide some context on this change.

We've gotten a lot of feedback from the Crypto Twitter community that the use of the term "Seed Phrase" was causing people to enter their MetaMask seed phrase into the DeSo Identity portal.

This criticism was not only accurate, but it also caused people in the ETH ecosystem to think that DeSo Identity was somehow trying to co-opt peoples' MetaMask wallets. I know it's clear to you and me that this isn't the case, but unfortunately the ubiquitous use of the term "Seed Phrase" in the ETH ecosystem caused confusion among the ETH community.

But I also agree with you guys that we shouldn't compromise on accuracy for the sake of reducing confusion if we can help it. So I propose that we call it "Private Key." This is actually what the seed phrase represents, and it's arguably *more* accurate to call it this as far as I can tell, since ultimately it is their private key that controls their account.

It also feels like this will have the positive side-effect of increasing the level of security with which people store their seed phrase because they know that it is PRIVATE and that it acts as the KEY to their account. I also like that it squares with what I tell people in conversation, which is that their account on DeSo is controlled by their private key.

So I am an approve with this change.
image

diamondhands0:
@triplegreenshell would you kindly integrate this change from my branch into yours pior to squashing+merging?

https://github.com/deso-protocol/identity/commits/triplegreenshell/DESO-KEY-COPY-UPDATE

diamondhands0 (approved):

tijno:
i really dont think this us any better and possibly more confusing.

a seed is not a private key

why not just add a note below:

> your deso seed is different from other coins like ETH, BTC

or 

> dont use your metamask seed

Bobnificent:
Not sure @diamondhands0 is aware that it was already changed to "sign up with DeSo Seed". Which makes it clear. If we turn it back to "Sign up with private key" It leads to a question... Which private key? Eth, BTC private key? I would suggest it stays "sign up with DeSo Seed" unless like @tijno says there is work being done to offer users a way to login with a "key" that is not a seed.

Screen Shot 2022-02-27 at 11 19 30

darian-parrish:
@diamondhands0 

I think a better consideration is just "login/signup with DeSo Identity" for the first frame. That sets the context as DeSo and will likely have the best longevity(also good branding everywhere).

Then on the next screen above the enter box(since the context is DeSo) can probably stick with Seed Phrase for now and maybe this becomes a derived key or something different in the future(but main CTA is "with DeSo Identity").

PR #123, opened 2022-03-28
Support DAO coin limit order transaction type

iamsofonias:
Identity changes to sign and submit DAO coin limit order tx type.

### DIFF
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -32,7 +32,8 @@ import {
   TransactionMetadataAuthorizeDerivedKey,
   TransactionMetadataMessagingGroup,
   TransactionMetadataDAOCoin,
-  TransactionMetadataTransferDAOCoin
+  TransactionMetadataTransferDAOCoin,
+  TransactionMetadataDAOCoinLimitOrder
 } from '../../lib/deso/transaction';
 import bs58check from 'bs58check';
 
@@ -247,6 +248,13 @@ export class ApproveComponent implements OnInit {
         const groupKeyName = this.transaction.metadata.messagingGroupKeyName
         description = `register group key with name "${groupKeyName}"`
         break
+      case TransactionMetadataDAOCoinLimitOrder:
+        const buyingDAOCoinCreatorPublicKey = this.base58KeyCheck(this.transaction.metadata.buyingDAOCoinCreatorPublicKey);
+        const sellingDAOCoinCreatorPublicKey = this.base58KeyCheck(this.transaction.metadata.sellingDAOCoinCreatorPublicKey);
+        const priceUnits = this.hexNanosToUnitString(this.transaction.metadata.priceNanos);
+        const quantityUnits = this.hexNanosToUnitString(this.transaction.metadata.quantityNanos);
+        description = `create DAO coin limit order to trade ${sellingDAOCoinCreatorPublicKey} coin ` +
+          `for ${buyingDAOCoinCreatorPublicKey} coin with price ${priceUnits} and quantity ${quantityUnits}`
     }
 
     // Set the transaction description based on the description populated with public keys.
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -249,12 +249,16 @@ export class ApproveComponent implements OnInit {
         description = `register group key with name "${groupKeyName}"`
         break
       case TransactionMetadataDAOCoinLimitOrder:
-        const buyingDAOCoinCreatorPublicKey = this.base58KeyCheck(this.transaction.metadata.buyingDAOCoinCreatorPublicKey);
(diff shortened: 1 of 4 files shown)


src/app/approve/approve.component.ts
```diff
@@ -247,6 +248,17 @@ export class ApproveComponent implements OnInit {
         const groupKeyName = this.transaction.metadata.messagingGroupKeyName
         description = `register group key with name "${groupKeyName}"`
         break
+      case TransactionMetadataDAOCoinLimitOrder:
+        let buyingDAOCoinCreatorPublicKey = this.base58KeyCheck(this.transaction.metadata.buyingDAOCoinCreatorPublicKey);
+        if (buyingDAOCoinCreatorPublicKey == "") {
+          buyingDAOCoinCreatorPublicKey = "DESO";
+        }
```
lazynina, line 255: If it's not the empty string, we'll want to add it to the `publicKeys` variable so we can get a username in place of the public key in the description in identity. See line 244 for how it's handled for DAO Coin transfers

src/app/approve/approve.component.ts
```diff
@@ -247,6 +248,17 @@ export class ApproveComponent implements OnInit {
...
+        let sellingDAOCoinCreatorPublicKey = this.base58KeyCheck(this.transaction.metadata.sellingDAOCoinCreatorPublicKey);
+        if (sellingDAOCoinCreatorPublicKey == "") {
+          sellingDAOCoinCreatorPublicKey = "DESO"
+        }
```
lazynina, line 259: same as above regarding substituting public keys for usernames.

src/lib/deso/transaction.ts
```diff
@@ -335,6 +335,26 @@ export class TransactionMetadataTransferDAOCoin extends BinaryRecord {
...
+  @Transcode(FixedBuffer(32))
+  scaledExchangeRateCoinsToSellPerCoinToBuy: Buffer = Buffer.alloc(0);
+
+  @Transcode(FixedBuffer(32))
+  quantityToBuyInBaseUnits: Buffer = Buffer.alloc(0);
+
+  @Transcode(VarBuffer)
+  cancelExistingOrder: boolean = false;
+
+  @Transcode(VarBuffer)
```
lazynina, line 354: I think we actually want to create a type for `DeSoInputsByTransactor` and make this `@Transacode(ArrayOf(DeSoInputsByTransactor))`

I think the `DeSoInputsByTransactor` type would look like

```
export class DeSoInputsByTransactor extends BinaryRecord {
  @Transcode(FixedBuffer(33))
  transactorPublicKey: Buffer = Buffer.alloc(0);

  @Transcode(ArrayOf(TransactionInput))
  inputs: TransactionInput[] = [];
}
```

src/app/approve/approve.component.ts
```diff
@@ -247,6 +248,17 @@ export class ApproveComponent implements OnInit {
...
+        description = `create DAO coin limit order to sell ${sellingDAOCoinCreatorPublicKey} ` +
+          `to buy ${buyingDAOCoinCreatorPublicKey}`
```
lazynina, line 261:
```suggestion
        description = `create DAO coin limit order to sell ${sellingDAOCoinCreatorPublicKey} ` +
          `to buy ${buyingDAOCoinCreatorPublicKey}`
```

We want this description to say `sell LazyNina DAO Coin` if it's not selling DESO. Can you make that adjustment? Additionally, it would be great if we could include the price and quantity. I know that's tricky, but we want to make it as clear as possible what the user is authorizing.

src/lib/deso/transaction.ts
```diff
@@ -335,6 +335,26 @@ export class TransactionMetadataTransferDAOCoin extends BinaryRecord {
...
+  sellingDAOCoinCreatorPublicKey: Buffer = Buffer.alloc(0);
+
+  @Transcode(FixedBuffer(32))
+  scaledExchangeRateCoinsToSellPerCoinToBuy: Buffer = Buffer.alloc(0);
+
+  @Transcode(FixedBuffer(32))
+  quantityToBuyInBaseUnits: Buffer = Buffer.alloc(0);
+
+  @Transcode(VarBuffer)
```
lazynina, line 351:
```suggestion
  @Transcode(Boolean)
```

src/app/approve/approve.component.ts
lazynina: Also, we need to update the description if this is cancelling an order.

lazynina: you can see how we parse hexNanoStrings to get quantity from line 241. Figuring out scaled exchange rate will be a bit trickier, but I believe in you :)

lazynina (changes requested):

src/lib/deso/transaction.ts
iamsofonias: Just for my understanding, what are all the reasons we need the to decode the tx metadata in identity? From what I can tell, the primary motivation is to produce a meaningful description of the tx for the user. Anything else?

PR #131, opened 2022-04-14
Revoke Derived key UI

lazynina:
If `deleteKey=true` and `publicKey` and `derivedPublicKey` are both set, then we display this simpler revoke screen.
image


src/app/derive/derive.component.html
```diff
@@ -75,6 +75,17 @@
 
 
   </div>
+  <div class="container home-container" *ngIf="deleteKey">
+    <div>
+      <div class="fs-18px mt-30px">
+        <span class="fs-underline"><b>You're about to revoke signing access for a derivedPublicKey</b></span><br/>
```
lazynina, line 81:
```suggestion
        <span class="fs-underline"><b>You're about to revoke signing access for a Derived Public Key</b></span><br/>
```

diamondhands0:
Looks good.

AdityaChaudhary0005:
Thanks for adding this. This was very must needed.

PR #132, opened 2022-04-15
Support DAO coin limit orders in identity

iamsofonias:
This adds support for limit order transaction in identity, and address comments from previous RFC PR (https://github.com/deso-protocol/identity/pull/123/files).

Test Plan

Cancelling an order
image

Submitted an order to sell 3 DAO coins for 4 $DESO each
image

Submitted an order to buy 1 DAO coin for 0.1 $DESO each
image

Submitting an order with unknown operation type w/ exchange rate of 0.1 and quantity of 1
image

### DIFF
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -282,7 +282,9 @@ export class ApproveComponent implements OnInit {
         break
       case TransactionMetadataDAOCoinLimitOrder:
         const daoCoinLimitOrderMetadata = this.transaction.metadata as TransactionMetadataDAOCoinLimitOrder;
-        if (daoCoinLimitOrderMetadata.cancelOrderID.length != 0) {
+        if (daoCoinLimitOrderMetadata.cancelOrderID != null &&
+          daoCoinLimitOrderMetadata.cancelOrderID.length != 0) {
+
           // The transaction is cancelling an existing limit order
           const orderId = daoCoinLimitOrderMetadata.cancelOrderID.toString('hex');
           description = `cancel DAO coin limit order with OrderID: ${orderId}`;
@@ -328,6 +330,7 @@ export class ApproveComponent implements OnInit {
             description = `create a DAO coin limit order to buy ${quantityToFill} ${buyingCoin} with an ` +
               `exchange rate of ${exchangeRate} ${sellingCoin} per coin bought`;
           }
+          // FIXME: You always need to catch the error case!
         }
         break;
     }
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -282,14 +282,12 @@ export class ApproveComponent implements OnInit {
         break
       case TransactionMetadataDAOCoinLimitOrder:
         const daoCoinLimitOrderMetadata = this.transaction.metadata as TransactionMetadataDAOCoinLimitOrder;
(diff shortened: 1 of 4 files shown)


lazynina (changes requested):
If you could add a couple screenshots of how the approval screen looks that'd be awesome

src/app/approve/approve.component.ts
```diff
@@ -276,9 +277,59 @@ export class ApproveComponent implements OnInit {
...
+          // The transaction is creating a new limit order
+          publicKeys = [];
+
+          let buyingCoin = '$DESO';
+          let sellingCoin = '$DESO';
+
+          // If the buying coin's public key is a zero byte array, then it means the is $DESO. Otherwise, it's a DAO coin
+          if (!this.isZeroByteArray(daoCoinLimitOrderMetadata.buyingDAOCoinCreatorPublicKey)) {
+            const buyingCoinName = this.base58KeyCheck(daoCoinLimitOrderMetadata.buyingDAOCoinCreatorPublicKey);
+            buyingCoin = buyingCoinName + ' coins';
```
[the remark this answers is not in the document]
iamsofonias: Yeah, 'DAO coins' seems to be consistent with the descriptions for other transactions too

lazynina (approved):

src/app/approve/approve.component.ts
```diff
@@ -301,9 +359,21 @@ export class ApproveComponent implements OnInit {
   }
 
   nanosToUnitString(nanos: number): string {
+    return this.toFixedLengthDecimalString(nanos / 1e9);
+  }
+
+  hexScaledExchangeRateToFloat(hex: Buffer): number {
+    return parseInt(hex.toString('hex'), 16) / 1e38;
```
AeonSw4n, line 366: doesn't this get crazy with 1e38? I remember we had to use BigInts to format nanos like in /lib/bindata/util.ts and that already was overflowing hard.

src/app/identity.service.ts
```diff
@@ -33,7 +33,7 @@ import {
   TransactionMetadataCreateNFT,
   TransactionMetadataDAOCoin,
   TransactionMetadataTransferDAOCoin,
-  TransactionSpendingLimit
```
AeonSw4n, line 36: Hmm, I wonder why this was here before and wasn't used?

AeonSw4n (approved):
Looks great! Just the 1e38 thing and a question about the removed import.

src/app/approve/approve.component.ts
lazynina: Yeah we should use the BN library for this parsing

src/app/identity.service.ts
lazynina: Diamondhands made some changes to use a hex string instead of transaction spending limits in this file and we didn't remove the imports.

src/app/approve/approve.component.ts
iamsofonias: Good point. Looks like the javascript number type [supports values up to ~1e308](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_VALUE), but starts to lose precision [beyond ~9e15](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger)

Guess, the precision loss was just not noticeable in my testing. I'll switch to bigints.

PR #133, opened 2022-04-19
Fix toFixedLengthDecimalString to remove trailing zeros

lazynina:
example of regex working: https://regex101.com/r/e36G1Q/1

previous version of regex on same test strings: https://regex101.com/r/6Dk0hH/1

### DIFF
--- a/src/app/approve/approve.component.ts
+++ b/src/app/approve/approve.component.ts
@@ -369,7 +369,7 @@ export class ApproveComponent implements OnInit {
   toFixedLengthDecimalString(num: number): string {
     // Change nanos into a formatted string of units. This combination of toFixed and regex removes trailing zeros.
     // If we do a regular toString(), some numbers can be represented in E notation which doesn't look as good.
-    return num.toFixed(9).replace(/([0-9]+(\.[0-9]+[1-9])?)(\.?0+$)/, '$1');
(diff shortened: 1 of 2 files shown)


src/app/approve/approve.component.ts
```diff
@@ -369,7 +369,7 @@ export class ApproveComponent implements OnInit {
   toFixedLengthDecimalString(num: number): string {
     // Change nanos into a formatted string of units. This combination of toFixed and regex removes trailing zeros.
     // If we do a regular toString(), some numbers can be represented in E notation which doesn't look as good.
-    return num.toFixed(9).replace(/([0-9]+(\.[0-9]+[1-9])?)(\.?0+$)/, '$1');
+    return num.toFixed(9).replace(/^(\d*\.\d*?[1-9])0+$/, '$1');
```
iamsofonias, line 372: Trying a few examples in console, it doesn't look like this gives us exactly what we want when dealing with integers:

Ex:
```
(10).toFixed(9).replace(/^(\d*\.\d*?[1-9])0+$/, '$1')
'10.000000000'
```

We should only apply the regex if the number is not an integer.

iamsofonias (changes requested):
Just one comment re: integers

lazynina:
> Just one comment re: integers

Check out the latest commit

iamsofonias (approved):

AeonSw4n (approved):
👍

PR #154, opened 2022-08-01
Derive call should only happen on derive route, fix derive flow when user needs to create an account

### LINKED ISSUE #139: "transactionSpendingLimitHex" is undefined when creating a derived key and user is not logged
1. Launch a window asking for derived keys permission:
```
    const transactionSpendingLimit =
    {
      "GlobalDESOLimit": 1000000000,
      "TransactionCountLimitMap":
      {
        "SUBMIT_POST": 1000000,
        "AUTHORIZE_DERIVED_KEY": 2
      }
    };

    identityWindow = window.open('https://identity.deso.org/derive?transactionSpendingLimitResponse=' + encodeURIComponent(JSON.stringify(transactionSpendingLimit)));
```
2. Click on "Log in with seed"
3. Enter the seed phrase of a user
4. Click on "Load Account"
5. In the returned payload, the "transactionSpendingLimitHex" is undefined, all other properties contain correct data:

```
function handleDerive(payload)
{
  if (identityWindow)
  {
    identityWindow.close();
    identityWindow = null;

    var publicKey = payload.publicKeyBase58Check;
    var derivedPublicKey = payload.derivedPublicKeyBase58Check;
    var derivedSeedHex = payload.derivedSeedHex;
    var expirationBlock = payload.expirationBlock;
    var accessSignature = payload.accessSignature;
    var transactionSpendingLimitHex = payload.transactionSpendingLimitHex;
```

If on step 2 you click on an already logged user, the payload contains the correct "transactionSpendingLimitHex" data

lazynina:
This will resolve #139 

Basically we weren't redirecting the user to the derive page after completing the sign-up flow with a new seed phrase. This update also redirects users to the derive page again from any flow. This ensures that the user explicitly approves a derived key - before a user could arrive at the derive page and then log into an existing account with a seed and that would automatically send the derive message back to the client.

### DIFF
--- a/src/app/sign-up/sign-up.component.ts
+++ b/src/app/sign-up/sign-up.component.ts
@@ -118,6 +118,11 @@ export class SignUpComponent implements OnInit, OnDestroy {
       this.router.navigate(['/', RouteNames.GET_DESO], {
         queryParams: {publicKey: this.publicKeyAdded, signedUp: true },
         queryParamsHandling: 'merge'});
+    } else if (this.globalVars.derive) {
+      this.globalVars.signedUp = true;
+      this.router.navigate(['/', RouteNames.DERIVE], {
+        queryParams: {publicKey: this.publicKeyAdded, signedUp: true},
+        queryParamsHandling: 'merge'});
     } else {
       this.identityService.login({
         users: this.accountService.getEncryptedUsers(),
--- a/src/app/account.service.ts
+++ b/src/app/account.service.ts
@@ -174,7 +174,8 @@ export class AccountService {
       messagingPrivateKey,
       messagingKeyName,
       messagingKeySignature,
-      transactionSpendingLimitHex
+      transactionSpendingLimitHex,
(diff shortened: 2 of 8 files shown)


src/app/sign-up/sign-up.component.ts
```diff
@@ -118,6 +118,11 @@ export class SignUpComponent implements OnInit, OnDestroy {
       this.router.navigate(['/', RouteNames.GET_DESO], {
         queryParams: {publicKey: this.publicKeyAdded, signedUp: true },
         queryParamsHandling: 'merge'});
+    } else if (this.globalVars.derive) {
```
AeonSw4n, line 121: `this.globalVars.derive` is set to true in the `derive.component` but also in `app.component`. Just want to triple-check that latter value write doesn't somehow overlap with this, but I think we're fine, right?

AeonSw4n (approved):
I liek

PR #176, opened 2022-09-21
Remove metamask on screens smaller than 576px

superzordon:


src/app/log-in-options/log-in-options.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="d-flex align-items-center flex-column">
   <button
-    class="button button-large mb-20px bg-dark text-white"
+    class="button button-large mb-20px bg-dark text-white d-none d-sm-flex"
```
DeSoDog: This will likely still show up on tablets and other larger mobile devices. If we wanted to guarantee no mobile device can access the metamask button we can use
https://www.npmjs.com/package/ngx-device-detector.

Which would look like the following:

constructor(deviceService: DeviceDetectorService) {
     
    }
  onInIt() {
this.mobile = this.deviceService.isMobile();
}

then in our template:

<button *ngIf="!mobile">...</button>

@lazynina @superzordon what are your thoughts?

DeSoDog (approved):

src/app/log-in-options/log-in-options.component.html
DeSoDog: Looks good

PR #190, opened 2022-09-27
Added Cookie for user

DeSoDog:


AeonSw4n (changes requested):
maybe we store PublicUserInfo instead?

PR #199, opened 2022-10-04
Ln/fix cookie usage

lazynina:


AeonSw4n (approved):
backward-compatibility is my only concern, though I got to say the code is much cleaner after adding `derivedPublicKeyBase58Check` and `encryptedMessagingKeyRandomness` to request/responses. Nice work!

PR #217, opened 2022-12-21
associations transcoders, access levels, and approve flow descriptions

lazynina:


src/environments/environment.ts
```diff
@@ -5,7 +5,7 @@
 export const environment = {
   production: false,
   hostname: 'https://identity.deso.org',
-  nodeURL: 'https://node.deso.org',
+  nodeURL: 'https://rt.deso.run',
```
lazynina, line 8: temporary

PR #218, opened 2023-01-06
feat: changes for new desojs redirect flow

jackson-dean:
This is to integrate with the derived key identity lib https://github.com/deso-protocol/deso-workspace/pull/91

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve --port 4201",
+    "start": "ng serve --port 4201 -c development",
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint",
@@ -106,5 +106,9 @@
     "ts-node": "~8.3.0",
     "tslint": "~6.1.0",
     "typescript": "~4.4.0"
+  },
+  "volta": {
+    "node": "16.19.0",
+    "npm": "7.24.2"
   }
 }
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve --port 4201 -c development",
+    "start": "ng serve --port 4201",
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -106,5 +106,9 @@
     "ts-node": "~8.3.0",
     "tslint": "~6.1.0",
     "typescript": "~4.4.0"
+  },
+  "volta": {
```
jackson-dean, line 110: this makes sure I'm running the correct versions of node and npm when I work on this. If you're not using this, it's pretty nice https://volta.sh

src/app/auth/google/google.component.ts
```diff
@@ -227,6 +227,7 @@ export const getStateParamsFromGoogle = (
     deleteKey: false,
     derivedPublicKey: '',
     expirationDays: 0,
+    redirect_uri: '',
```
lazynina, line 230: we're also going to need to update `google-drive.service.ts` around line 97 to add `redirectUri: this.globalVars.redirectUri` so it gets fed into the google flow and we still have the param when we get back from the google flow.

src/app/log-in/log-in.component.html
```diff
@@ -2,7 +2,7 @@
 
 <div
   class="container home-container text-center"
-  *ngIf="globalVars.inTab || globalVars.webview"
+  *ngIf="globalVars.inTab || globalVars.webview || globalVars.redirectURI"
```
lazynina, line 5: is there a reason we don't have `|| globalVars.callback` in here?

lazynina (changes requested):

src/app/auth/google/google.component.ts
jackson-dean: damn good catch, yeah i tried the google flow with popup but did not do the full page redirect + google flow. Just test and confirmed it was broken and this fixes it.

src/app/log-in/log-in.component.html
jackson-dean: So my theory is that callback is really only used for mobile app use cases, and I believe mobile apps pass the webview param as well, so for the purpose of some of these template conditions I *think* they are interchangeable.

lazynina (approved):

PR #225, opened 2023-02-01
fix derive login -> get deso flow

jackson-dean:
Selecting an account shouldn't skip the get free money flow if the balance is zero.


src/app/derive/derive.component.ts
```diff
@@ -150,6 +154,28 @@ export class DeriveComponent implements OnInit {
...
   }
 
+  onAccountSelected(publicKey: string): void {
+    // first check if this user has a balance. If not, push them to the get-deso flow
+    this.backendApi
+      .GetUsersStateless(
+        [publicKey],
+        true /*SkipForLeaderboard*/,
+        true /*IncludeBalance*/,
+        true /*GetUnminedBalance*/
```
lazynina, line 164:
```suggestion
        true /*IncludeBalance*/
```

lazynina (approved):

PR #227, opened 2023-02-02
display count values >= 1e9 as UNLIMITED

jackson-dean:
It's a common desire to want to set individual tx types to unlimited. We currently only have the ability to make all transactions unlimited or specify an explicit number per tx type. This makes it such that any value greater than or equal to 1 billion will get displayed as "UNLIMITED." This also matches the unlimited tx type we have implemented in [@deso-core/identity](https://github.com/deso-protocol/deso-workspace/blob/c517a1c64f75afcd3e97a8ae383a6e92131aa956/libs/identity/src/lib/permissions-utils.spec.ts#L6).

Screenshot 2023-02-02 at 1 24 50 PM

### DIFF
--- a/src/app/global-vars.service.ts
+++ b/src/app/global-vars.service.ts
@@ -160,7 +160,7 @@ export class GlobalVarsService {
   }
 
   // If the count is 1 billion or more, it will be displayed as "UNLIMITED"
-  // Otherwise, it will be displayed as a delimited number based on the users locale.
+  // Otherwise, it will be displayed as a delimited number based on the user's locale.
   formatTxCountLimit(count: number = 0): string {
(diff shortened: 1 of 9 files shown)


src/app/global-vars.service.ts
```diff
@@ -158,4 +158,10 @@ export class GlobalVarsService {
       )
       .join(' ');
   }
+
+  // If the count is 1 billion or more, it will be displayed as "UNLIMITED"
+  // Otherwise, it will be displayed as a delimited number based on the users locale.
```
jackson-dean, line 163:
```suggestion
  // Otherwise, it will be displayed as a delimited number based on the user's locale.
```

lazynina (approved):
small copy changes, but otherwise approved.

PR #231, opened 2023-02-13
instrument interaction events

jackson-dean:


src/app/buy-deso/buy-deso-megaswap/buy-deso-megaswap.component.html
```diff
@@ -11,6 +11,7 @@
   </ng-template>
   <div class="d-flex justify-content-center">
     <button
+      data-control-name="buy-deso-finish-flow-button"
```
lazynina, line 14: I think we've got a public key at this point - should we log the public key of the user who finished the flow?

src/app/buy-or-send-deso/buy-or-send-deso.component.html
```diff
@@ -31,6 +31,7 @@
           <div class="d-flex justify-content-around">
             <div class="col-10 d-flex align-items-center flex-column">
               <button
+                data-control-name="buy-or-send-buy-now-button"
```
lazynina, line 34: I won't comment on each of these, but I think everywhere we have a public key, we should log it in the interaction event.

lazynina (commented):
Look good other than logging public keys where possible.

src/app/buy-deso/buy-deso-megaswap/buy-deso-megaswap.component.html
jackson-dean: yep i added a thing to just always pull it from the query params. if it's not there yet it'll just be empty which seems fine. The whole stream of interactions should be back associated once diamond identifies the user after login.

src/app/buy-or-send-deso/buy-or-send-deso.component.html
jackson-dean: I think the global query param parser thing should take care of this.

src/app/backend-api.service.ts
```diff
@@ -288,6 +289,13 @@ export class BackendAPIService {
 
   jwtPost(path: string, publicKey: string, body: any): Observable<any> {
     const publicUserInfo = this.accountService.getEncryptedUsers()[publicKey];
+    // NOTE: there are some cases were derived user's were not being sent phone number
```
lazynina, line 292:
```suggestion
    // NOTE: there are some cases where derived user's were not being sent phone number
```

lazynina (approved):

PR #237, opened 2023-06-09
skip derived key approval page if the user has no money

jackson-dean:


jackson-dean:
We actually don't need (or want) to do this.

PR #242, opened 2023-07-06
upgrade from wallet connect 1.x -> 2.x

jackson-dean:
A couple of interesting improvements from 1.x:
- We now have access to an ethereum provider that can be passed to ethersjs such that it has the same API as the desktop extension which reduces a lot of code branching.
- We are provided the deeplink for metamask *directly* which we can open, instead of needing the intermediate wallet connect QR modal popup.
- Overall the mobile experience is now quite similar to the desktop extension experience.


src/app/account.service.ts
```diff
@@ -80,7 +81,10 @@ export class AccountService {
       );
       let encryptedMessagingKeyRandomness: string | undefined;
       if (privateUser.messagingKeyRandomness) {
-        encryptedMessagingKeyRandomness = this.cryptoService.encryptSeedHex(privateUser.messagingKeyRandomness, hostname);
+        encryptedMessagingKeyRandomness = this.cryptoService.encryptSeedHex(
```
jackson-dean, line 84: just fyi, a lot of this (like this change) is just formatting noise from formatting the file, which is annoying, but it's not a real change.

lazynina (approved):

PR #263, opened 2023-09-20
chore: update configs only

jackson-dean:
This updates some configs for development.
- Fix the configuration for husky to auto format code when it is committed so we don't get into state of bad formatting causing a lot of unnecessary PR noise.
- Add CI quality check in case someone bypasses local precommit hook using `-n` flag.
- We pass the `development` configuration to the
dev server start script because it compiles faster.
- Update the tsconfig lib to es2020 to enable syntax
that is now widely supported, specifically Array.prototype.flat
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

PR #262, opened 2023-09-20
chore: code formatting only

jackson-dean:
This is just a preliminary PR that formats the code prior to making any
functional changes. This only running prettier across some of the files and
nothing more.

PR #264, opened 2023-09-20
refactor: update account service to support sub accounts

jackson-dean:
Changes the way accounts are stored to support sub accounts.
An overview of the storage strategy:
- We continue to store "root" accounts (accounts derived with account number 0)
in a `users` key in local storage that is keyed on the public key of the root account.
- All subaccounts are stored in a `subAccounts` property on the root account. These subaccounts
consist of the following props: `{ accountNumber, lastLoginTimestamp, isHidden }`. The public key
can be derived from this information combined with the seed stored on the root user.
- We have a reverse lookup map that maps subaccount public keys back to their root account public key.
We need this to support look up by public key when a consuming application passes us a public key for
the logged in user.

NOTE: There is nothing special about the root account. It is only for historical reasons that we key
the common data (seed, mnemonic, etc) under this account's public key. In practice, all accounts are
derived the same and all sub accounts are fully fledged and unique "owner" accounts that can be
issued their own derived keys and can sign with their own public/private keys.


jackson-dean:
Current dependencies on/for this PR:
* main
  * **PR #262** <a href="https://app.graphite.dev/github/pr/deso-protocol/identity/262" target="_blank">Graphite</a> 
    * **PR #263** <a href="https://app.graphite.dev/github/pr/deso-protocol/identity/263" target="_blank">Graphite</a> 
      * **PR #264** <a href="https://app.graphite.dev/github/pr/deso-protocol/identity/264" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/deso-protocol/identity/264?utm_source=stack-comment).

lazynina (approved):

PR #278, opened 2023-09-23
DELETEME squash

diamondhands0:


src/lib/account-number.ts
```diff
@@ -0,0 +1,53 @@
...
+    return candidate;
+  }
+
+  // At most we look back 500 numbers. This is a bit arbitrary...  but the
+  // number of values could *technically* be 2^32 - 1, so we just limit the
+  // number of iterations to some reasonable value. The reason we look back for
+  // the highest available number instead of picking the lowest number is that
+  // the lowest number is more likely to have been used in the past and we're
+  // aiming to get a fresh wallet.
+  const maxLookBack = Math.max(sorted.length - 500, 0);
```
diamondhands0, line 43: How would we ever hit this? Looks like it can only happen if candidate = MAX_UNSIGNED_INT_VALUE, in which case maybe we should just return 0 or some default value.

src/app/account.service.ts
```diff
@@ -38,6 +39,12 @@ import { MetamaskService } from './metamask.service';
 import { SigningService } from './signing.service';
 
 export const ERROR_USER_NOT_FOUND = 'User not found';
+const SUB_ACCOUNT_REVERSE_LOOKUP_KEY = 'subAccountReverseLookup';
+
+export interface SubAccountReversLookupEntry {
```
diamondhands0, line 44: Add comments around this.

src/app/account.service.ts
```diff
@@ -51,21 +58,89 @@ export class AccountService {
   constructor(
     private cryptoService: CryptoService,
     private globalVars: GlobalVarsService,
-    private cookieService: CookieService,
     private entropyService: EntropyService,
     private signingService: SigningService,
     private metamaskService: MetamaskService
-  ) {}
+  ) {
+    this.initializeSubAccountReverseLookup();
```
diamondhands0, line 65: Add comments around this.

src/app/crypto.service.ts
```diff
@@ -279,3 +303,22 @@ export class CryptoService {
     return ethAddressChecksum;
   }
 }
+
+/**
+ * We set the account according to the following derivation path scheme:
+ * m / purpose' / coin_type' / account' / change / address_index
+ * See for more details: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account
+ */
+function deriveKeys(
```
diamondhands0, line 312: Can we call these "subaccountKeys" rather than "derived" keys to avoid overloading/confusion? We can call this function generateSubaccountKeys() maybe.

src/lib/account-number.ts
jackson-dean: ahh, the way you can hit this is if you enter the max value directly into the recover thing. It's not likely, but possible. If that happens and you return just a default value, like zero, then you hit collisions with existing accounts, so we do need to do some search fora free spot. The question is just whether we search for the lowest or the highest value. I felt like searching for the highest value available makes the most sense bc it's far less likely that it would hit a recoverable account that you've already used before.

src/app/crypto.service.ts
jackson-dean: yeah, i agree it's a little confusing. Will update in the other PR 👍🏻

src/app/grouped-account-select/grouped-account-select.component.ts
```diff
@@ -0,0 +1,311 @@
...
+    });
+  }
+
+  addSubAccount(
+    rootPublicKey: string,
+    { accountNumber }: { accountNumber?: number } = {}
+  ) {
+    const addedAccountNumber = this.accountService.addSubAccount(
+      rootPublicKey,
+      { accountNumber }
```
jackson-dean, line 227: just a note for myself, I we should change this to make accountNumber a required argument here.

src/app/grouped-account-select/grouped-account-select.component.ts
```diff
@@ -0,0 +1,311 @@
...
+    const rootAccount = this.accountService.getAccountInfo(rootPublicKey);
+    return !rootAccount.exportDisabled;
+  }
+
+  exportSeed(rootPublicKey: string) {
+    const dialogRef = this.dialog.open(BackupSeedDialogComponent, {
+      data: { rootPublicKey },
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
```
jackson-dean, line 307: delete this.

